PR #20, opened 2023-11-14
View name override

rkistner:
Supports:
1. Overriding the view name for tables: `Table(..., viewName: 'local_table')`.
2. Updating the schema after the database has already been opened (but before connecting sync): `await db.updateSchema(schema)`

This allows having separate tables for a synced mode and a local-only mode, and moving data between the two - see the tests. The use case is a user first using the app completely offline (without registering an account), then potentially at a later stage connecting an account and syncing data, without the upload queue continuously growing in size before that.

Additional changes:
1. Validate schema definitions for duplicates and invalid names.
2. Support preserving local-only tables when clearing database: `await db.disconnectAndClear(clearLocal: false)`.


kobiebotha (approved):
🚀

PR #24, opened 2023-12-20
[Chore] Move to monorepo with attachments SDK

DominicGBauer:
## Description
Implement [melos](https://melos.invertase.dev/) as the monorepo tool of choice and move select flutter packages and demos into a monorepo.

## Work done
* Updated CI/CD to run based on changes to packages or demos in repo
* Added packages subdirectory to house all dart/flutter packages we create (right not it contains powersync and powersync_attachments_helper)
* Added a demos subdirectory to house all dart/flutter demos we create

## How to test
The easiest way would be to run the demo flutter todolist app and make sure it still works as expected


rkistner (commented):
Nitpick: Can the package folders be renamed to just `powersync` and `powersync_attachments_helper`, without the `.dart`?

rkistner (approved):

PR #41, opened 2024-01-02
Handle missing camera in demo

rkistner:
The attachments demo uses the camera, which is only available on Android and iOS.

This changes the camera setup to only run when taking a photo, and display an error if it cannot be initialized (e.g. on desktop).

image

### DIFF
--- a/demos/supabase-todolist/lib/attachments/photo_widget.dart
+++ b/demos/supabase-todolist/lib/attachments/photo_widget.dart
@@ -38,7 +38,7 @@ class _PhotoWidgetState extends State<PhotoWidget> {
 
     bool fileExists = await File(photoPath).exists();
 
-    return _ResolvedPhotoState(photoPath: null, fileExists: fileExists);
+    return _ResolvedPhotoState(photoPath: photoPath, fileExists: fileExists);
   }
 
   @override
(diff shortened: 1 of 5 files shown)


demos/supabase-todolist/lib/attachments/photo_widget.dart
```diff
@@ -54,29 +82,25 @@ class _PhotoWidgetState extends State<PhotoWidget> {
...
-            Key key = ObjectKey('$filePath:$lastModified');
-
-            return Image.file(
-              key: key,
-              imageFile,
-              width: 50,
-              height: 50,
-            );
+          String? filePath = data.photoPath;
+          bool fileIsDownloading = data.fileExists;
```
DominicGBauer, line 86: This should be `!data.fileExists`

demos/supabase-todolist/lib/attachments/photo_widget.dart
```diff
@@ -19,31 +20,58 @@ class PhotoWidget extends StatefulWidget {
...
     if (photoId == null) {
-      return {"photoPath": null, "fileExists": false};
+      return _ResolvedPhotoState(photoPath: null, fileExists: false);
     }
     photoPath = await attachmentQueue.getLocalUri('$photoId.jpg');
 
     bool fileExists = await File(photoPath).exists();
 
-    return {"photoPath": photoPath, "fileExists": fileExists};
+    return _ResolvedPhotoState(photoPath: null, fileExists: fileExists);
```
DominicGBauer, line 41: this should return the photoPath and not null

DominicGBauer (changes requested):

demos/supabase-todolist/lib/attachments/photo_widget.dart
rkistner: Thanks, fixed.

DominicGBauer (approved):

PR #42, opened 2024-01-02
Improve connection status

rkistner:
Changes to SyncStatus:
 * Fix `connected` not updating when calling `powersync.disconnect()`.
 * Add `connecting`, `downloading` and `uploading` flags.
 * Add `downloadError`, `uploadError` and a combined `exception` field.

The demo now uses these to provide more fine-grained connectivity status.

image

Also includes a minor fix for error parsing that would only return the first character of the error message in some cases.

### DIFF
--- a/demos/supabase-anonymous-auth/lib/widgets/status_app_bar.dart
+++ b/demos/supabase-anonymous-auth/lib/widgets/status_app_bar.dart
@@ -61,13 +61,13 @@ Widget _makeIcon(String text, IconData icon) {
 }
 
 Widget _getStatusIcon(SyncStatus status) {
-  if (status.exception != null) {
-    // The exception message is verbose, could be replaced with something
+  if (status.anyError != null) {
+    // The error message is verbose, could be replaced with something
     // more user-friendly
     if (!status.connected) {
-      return _makeIcon(status.exception!.toString(), Icons.cloud_off);
+      return _makeIcon(status.anyError!.toString(), Icons.cloud_off);
     } else {
-      return _makeIcon(status.exception!.toString(), Icons.sync_problem);
+      return _makeIcon(status.anyError!.toString(), Icons.sync_problem);
     }
   } else if (status.connecting) {
     return _makeIcon('Connecting', Icons.cloud_sync_outlined);
--- a/demos/supabase-anonymous-auth/lib/widgets/status_app_bar.dart
+++ b/demos/supabase-anonymous-auth/lib/widgets/status_app_bar.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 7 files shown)


packages/powersync/lib/src/sync_status.dart
```diff
@@ -2,27 +2,72 @@ class SyncStatus {
   /// true if currently connected
   final bool connected;
 
+  /// true if currently connected
```
kobiebotha, line 5: should this be "true if establishing a connection" ?

demos/supabase-anonymous-auth/lib/widgets/status_app_bar.dart
```diff
@@ -39,22 +40,48 @@ class _StatusAppBarState extends State<StatusAppBar> {
...
   }
 }
+
+Widget _makeIcon(String text, IconData icon) {
+  return Tooltip(
+      message: text,
+      child: SizedBox(width: 40, height: null, child: Icon(icon, size: 24)));
+}
+
+Widget _getStatusIcon(SyncStatus status) {
```
DominicGBauer, line 63: I think a switch statement would be nicer here

demos/supabase-anonymous-auth/lib/widgets/status_app_bar.dart
```diff
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:flutter/foundation.dart';
```
DominicGBauer, line 3: Based on all the repeated changes should we not just create a common component for that status bar and then have all the demos use it?

demos/supabase-anonymous-auth/lib/widgets/status_app_bar.dart
rkistner: A switch statement can't quite handle all the conditions used here - at least not without being more verbose.

demos/supabase-anonymous-auth/lib/widgets/status_app_bar.dart
rkistner: Technically the status bar is slightly different between demos (one contains search), but the status icon is the same and could be re-used.

I'm a little conflicted on this one: One the one side a common package would allow us to re-use code between the demos, but on the other side it makes it more difficult for others to copy and modify a demo.

What I think would actually be best is to reduce the scope of some of the demos (specifically supabase-anonymous-auth and supabase-edge-auth) to not contain the entire todolist app, but rather just a minimal amount of code to show how to use custom auth functions.

DominicGBauer: Yeah that makes sense. I left the status bar in the auth demos because of the status icon mainly but don't think it's really necessary

kobiebotha (approved):

DominicGBauer (approved):

PR #45, opened 2024-01-04
Add return type to PowerSyncDatbase connect

thedalelakes:
Adds missing return type to this async library function


rkistner (approved):
Thanks, will include in the next release.

PR #53, opened 2024-01-24
Improved error logging

rkistner:
This changes the logging behavior to log to console in debug mode by default, instead of logging to the root Logger.

Additionally, error messages are somewhat improved for cases where the error itself is not logged.

To get the old logging behavior back for full control over logging, use `logger: attachedLogger` in the PowerSyncDatabase constructor.

Breaking changes:
1. Exceptions are different in certain conditions.
2. Default logging behavior changed as described above.
3. PowerSyncCredentials does not have a const constructor anymore, and now validates the endpoint URI.

Despite these being breaking changes, I believe it will make debugging significantly easier for new apps, and not impact existing apps significantly.

The example apps still contain the same log handling code, since there are other places in the examples also using the logger. However, it is not required to get debug output for the powersync lib anymore.

### DIFF
--- a/packages/powersync/lib/src/powersync_database.dart
+++ b/packages/powersync/lib/src/powersync_database.dart
@@ -68,7 +68,11 @@ class PowerSyncDatabase with SqliteQueries implements SqliteConnection {
   /// null when disconnected, present when connecting or connected
   AbortController? _disconnecter;
 
-  /// The Logger internally used by this PowerSyncDatabase
+  /// The Logger used by this [PowerSyncDatabase].
+  ///
+  /// The default is [autoLogger], which logs to the console in debug builds.
+  /// Use [debugLogger] to always log to the console.
+  /// Use [attachedLogger] to propagate logs to [Logger.root] for custom logging.
   late final Logger logger;
 
   /// Open a [PowerSyncDatabase].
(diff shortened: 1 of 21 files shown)


packages/powersync/lib/src/powersync_database.dart
```diff
@@ -602,3 +643,15 @@ Future<void> _powerSyncDatabaseIsolate(
     throw error;
   });
 }
+
+enum LogType {
```
DominicGBauer, line 647: this should probably be in [packages/powersync/lib/src/log.dart](https://github.com/powersync-ja/powersync.dart/pull/53/files#diff-91cae18e041640056785a3479792bd713b26429c83ecce1bc1236a8474018288)

packages/powersync/lib/src/streaming_sync.dart
```diff
@@ -386,3 +393,63 @@ String? stringOrFirst(Object? details) {
     return null;
   }
 }
+
+/// Attempt to give a basic summary of the error for cases where the full error
```
DominicGBauer, line 397: should we not move all of this into it's own file (including classes below)?

packages/powersync/lib/src/powersync_database.dart
```diff
@@ -98,18 +102,55 @@ class PowerSyncDatabase with SqliteQueries implements SqliteConnection {
...
   ///
   /// Migrations are run on the database when this constructor is called.
-  PowerSyncDatabase.withDatabase(
-      {required this.schema, required this.database}) {
+  PowerSyncDatabase.withDatabase({
+    required this.schema,
+    required this.database,
+    LogType log = LogType.auto,
+  }) {
+    if (log == LogType.debug || log == LogType.auto) {
```
DominicGBauer, line 124: can we extract this into a method that returns the logger?

rkistner: I'm thinking it might be even better to pass in a Logger instance rather than the enum - will check if that's feasible

packages/powersync/lib/src/streaming_sync.dart
rkistner: I'm keeping this method here since it's private and only used in this one place, but I moved the classes to a separate exported file now.

DominicGBauer (approved):

PR #56, opened 2024-01-30
[Wip Alpha] Web Support

stevensJourney:
# Desciption

This adds initial support for Flutter Web. This extends the work done here https://github.com/powersync-ja/sqlite_async.dart/pull/25

# Status

## Testing
The SDK was tested using the Todo list demo app on Web, Android, iOS and MacOs platforms. 

https://github.com/powersync-ja/powersync.dart/assets/51082125/5b14f9bf-5a3d-4575-b08b-aad45dd587a5

### How to test
Initialise the repo with `melos prepare`

Run the `supabase-todolist` demo.

```
cd demos/supabase-todolist
flutter run -d chrome
```

## Tests
Currently chrome tests suffer from intermittent asset server issues which cause tests to fail. These tests do pass locally
image

Test will be fixed in a later alpha release.

Update 2024/02/12: Mosts tests now pass on Chrome in CI
image

TODOs:
- [x] Update Drift dependency
- [x] Bump `sqlite_async.dart` dependency once packages are available


melos.yaml
```diff
@@ -29,9 +31,19 @@ scripts:
...
+    packageFilters:
+      scope:
+        - powersync_web_worker
+
+  update:wasm: sh scripts/init_sqlite_wasm.sh
+
   test:
     description: Run tests in a specific package.
-    run: dart test
+    # Tests on Web are disabled for now. They pass locally but there is a network issue on CI
```
rkistner, line 45: Is this still accurate?

packages/powersync/lib/src/bucket_storage.dart
```diff
@@ -300,7 +303,7 @@ class BucketStorage {
       // |--SEARCH r USING INDEX ps_oplog_by_row (row_type=? AND row_id=?)
       // `--USE TEMP B-TREE FOR GROUP BY
       // language=DbSqlite
-      var stmt = db.prepare(
+      var opRows = await tx.execute(
```
rkistner, line 306: This change can have a big impact on memory usage. The previous version did it in batches - never keeping more than 10k rows in memory at a time (which could actually be reduced to a much smaller number / replaced saving individual rows at a time). This version keeps the entire resultset in memory, which could be massive for some use cases (100k+ rows).

Is it feasible to have separate implementations for web vs native as a stop-gap here, until we move to using the sqlite extension?

packages/powersync/lib/src/uuid.dart
```diff
@@ -2,4 +2,6 @@ import 'package:uuid/uuid.dart';
 import 'package:uuid/data.dart';
 import 'package:uuid/rng.dart';
 
-final uuid = Uuid(goptions: GlobalOptions(CryptoRNG()));
+import 'log_internal.dart';
+
+final uuid = kIsWeb ? Uuid() : Uuid(goptions: GlobalOptions(CryptoRNG()));
```
rkistner, line 7: Why not CryptoRNG on web?

packages/powersync/test/schema_test.dart
```diff
@@ -106,36 +109,39 @@ void main() {
...
-      await powersync.updateSchema(schema2);
-
-      // Execute instead of getAll so that we don't get a cached query plan
-      // from a different connection
-      final results2 = await powersync.execute(
-          'EXPLAIN QUERY PLAN SELECT * FROM assets WHERE make = ?', ['test']);
-
-      expect(results2[0]['detail'], contains('SCAN'));
-    });
+    /// TODO this test breaks on web at the moment.
```
rkistner, line 112: We should figure out why this is breaking. While this specific scenario may not be common in real apps, there could be an underlying issue that causes more weird bugs.

melos.yaml
stevensJourney: No, this is not accurate. It will be removed. Thanks for spotting!

packages/powersync/lib/src/uuid.dart
stevensJourney: Initially I thought this caused compiler errors, but I was not able to reproduce this now - my initial suspicions must have been caused by something else. Reverted this change to use `CryptoRNG` for all platforms.

packages/powersync/lib/src/bucket_storage.dart
stevensJourney: Yes, it is feasible. I've now implemented a `NativeBucketStorage` class which extends `BucketStorage`. This uses the sqlite connection directly to apply bucket changes.

packages/powersync/test/schema_test.dart
stevensJourney: Thanks for the investigation! This will be fixed in a future alpha release of `sqlite_async.dart`

stevensJourney: This test now passes after updates to the Drift worker

.github/workflows/demos.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Install Melos
         run: flutter pub global activate melos
       - name: Install dependencies
-        run: melos bootstrap
+        run: melos prepare
```
DominicGBauer, line 26: I don't think this command exists ? 
image

stevensJourney: It exists after being defined [here](https://github.com/powersync-ja/powersync.dart/blob/25a75f0cb5372381628f2990c9cbde2c62163758/melos.yaml#L12)

rkistner (approved):

PR #58, opened 2024-02-07
More detailed README and small bugfix

ckritzinger:
This adds quite a lot of detail to the README, I think it's basically everything that I would have needed to get Hello World up and running.

I'm not sure how you feel about about having a longer readme versus moving more detail to examples/. Personally I find it quite useful to have the main gist of things in a Readme that I can read from top to bottom.

I also fixed a small bug in `getExpiryDate` that I ran into. JWTs don't use padding but it seems that [dart:convert requires padding even for base64Url encoding](https://api.flutter.dev/flutter/dart-convert/base64Url-constant.html).

### DIFF
--- a/packages/powersync/README.md
+++ b/packages/powersync/README.md
@@ -11,48 +11,138 @@
 * No need for client-side database migrations - these are handled automatically.
 * Subscribe to queries for live updates.
 
+## Examples
+
+For a number of complete app examples, see our [example app gallery](https://docs.powersync.com/resources/demo-apps-example-projects#flutter)
+
 ## Getting started
+
+You'll need to create a PowerSync account and set up a PowerSync instance. You can do this at [https://www.powersync.com/](https://www.powersync.com/).
+
+### Install the package
+
+`flutter pub add powersync`
+
+### Implement a backend connector and initialize the Powersync database
 
 ```dart
 import 'package:powersync/powersync.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:path/path.dart';
 
+// Define the schema for the local SQLite database.
+// You can automatically generate this schema based on your sync rules:
+// In the Powersync dashboard, right-click on your PowerSync instance and then click "Generate client-side schema"
 const schema = Schema([
   Table('customers', [Column.text('name'), Column.text('email')])
 ]);
 
 late PowerSyncDatabase db;
 
-// Setup connector to backend if you would like to sync data.
-class BackendConnector extends PowerSyncBackendConnector {
+// You must implement a backend connector to define how Powersync communicates with your backend.
+class MyBackendConnector extends PowerSyncBackendConnector {
   PowerSyncDatabase db;
 
-  BackendConnector(this.db);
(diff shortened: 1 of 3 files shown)


packages/powersync/README.md
```diff
@@ -11,48 +11,141 @@
...
+// In the Powersync dashboard, right-click on your PowerSync instance and then click "Generate client-side schema"
 const schema = Schema([
   Table('customers', [Column.text('name'), Column.text('email')])
 ]);
 
 late PowerSyncDatabase db;
 
-// Setup connector to backend if you would like to sync data.
-class BackendConnector extends PowerSyncBackendConnector {
+// You must implement a backend connector to define how Powersync communicates with your backend.
```
benitav, line 44:
```suggestion
// You must implement a backend connector to define how PowerSync communicates with your backend.
```

ckritzinger: @benitav I'm going to correct all the examples of that in my changes, but note that there are a number of others sprinkled throughout the codebase.

packages/powersync/lib/src/connector.dart
```diff
@@ -113,7 +113,12 @@ class PowerSyncCredentials {
     try {
       List<String> parts = token.split('.');
       if (parts.length == 3) {
-        final rawData = base64Decode(parts[1]);
+        String part = parts[1];
+        // dart:convert doesn't like missing padding
+        if (part.length % 4 > 0) {
+          part += '=' * (4 - part.length % 4);
+        }
+        final rawData = base64Decode(part);
```
rkistner, line 121: The issue is that JWTs use base64-url encoding instead of standard base64. Apart from the padding, there are also other minor encoding differences. Could you check if this works?

```suggestion
        final rawData = base64Url.decode(part);
```

packages/powersync/README.md
benitav: Sounds good - I just happened to spot this one :)

packages/powersync/lib/src/connector.dart
ckritzinger: I'll test, but their docs suggest otherwise: https://api.flutter.dev/flutter/dart-convert/base64Url-constant.html

ckritzinger: I tested with your suggested change and still get the same FormatException about invalid length.

https://github.com/dart-lang/sdk/issues/39510 seems to suggest this is deliberate.

ckritzinger: However, there does seem to be a better way to do this, brb

ckritzinger: using https://api.flutter.dev/flutter/dart-convert/Base64Codec/normalize.html seems like a better approach

rkistner: Thanks! Turns out my initial suggested change doesn't even make a difference - `base64url.decode` and `base64.decode` does the same thing.

rkistner (approved):

PR #59, opened 2024-02-14
chore: remove devconnector references

DominicGBauer:
## Description
Remove references to DevConnector


rkistner (commented):
Completely removing it is a breaking change - rather just mark as deprecated and remove comments mentioning it. We can then remove when we do a v2.0 release.

rkistner (approved):

PR #62, opened 2024-02-15
[Web Alpha] Improvements 1

stevensJourney:
This adds some improvements after the initial alpha release.

- This updates the alpha-releases branch with the latest changes from `master`
- `PowerSyncDatabase` initialisation has been updated in README and example apps with better web compatibility. This removes any direct usage of `dart:io` and replaces some usages with `universal_io`. 
- Package README is updated to point to Github releases for `powersync_db.worker.js` file. Now also includes more user-friendly instructions.
- The SQLite3 WASM init script has been migrated to a Dart script for better dev setup compatibility. 
- Optimised `powersync_db.worker.js` compiled file size from 2Mb to 430kB.
- Updated Pana package analysing to only analyse published packages
- Added CI action for creating draft releases


demos/supabase-anonymous-auth/lib/powersync.dart
```diff
@@ -137,8 +136,13 @@ String? getUserId() {
 }
 
 Future<String> getDatabasePath() async {
-  final dir = kIsWeb ? Directory('/') : await getApplicationSupportDirectory();
-  return join(dir.path, 'powersync-demo.db');
+  const dbFilename = 'powersync-demo.db';
+  // getApplicationSupportDirectory is not supported on Web
+  if (kIsWeb) {
```
DominicGBauer, line 141: what is kIsWeb? Can this be named more clearly?

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+    tags:
+      - 'v*' # Trigger on tags beginning with 'v'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
```
DominicGBauer, line 15: I think v4 is latest

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install Melos
+        run: flutter pub global activate melos
+
+      - name: Install dependencies and compile assets
+        run: melos prepare
+
+      # Create the Draft release using the tag name
+      - name: Create Draft Release
+        id: create_release
+        uses: actions/create-release@v1
```
DominicGBauer, line 32: It looks like this was archived in 2021 perhaps we should rather use one of the ones mentioned in the archived repo?

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: true
+          prerelease: true # TODO set this as false in future
+
+      # Upload the compiled assets to the Draft release.
```
DominicGBauer, line 41: comment not required since name repeats

demos/supabase-anonymous-auth/lib/powersync.dart
stevensJourney: This is a standard import from Flutter. The naming comes from them.

.github/workflows/release.yml
stevensJourney: Good catch! I updated this to a newer action that does both release creation and asset uploading.

.github/workflows/release.yml
stevensJourney: True

PR #65, opened 2024-02-21
feat(attachments): add error handlers

DominicGBauer:
## Description
Add error handlers users can tap into

## Work Done
* Added upload and download error handlers users can use
* Error handler will ignore the attachment if the user opts for this
* Updated the demo to represent this

## How to test
* Create a todo and upload an image
* Delete the app
* Delete the image from local storage
* Reinstall the app
* You should see that the image has been ignored

### DIFF
--- a/packages/powersync_attachments_helper/lib/src/attachments_queue_table.dart
+++ b/packages/powersync_attachments_helper/lib/src/attachments_queue_table.dart
@@ -64,7 +64,7 @@ class Attachment {
 /// 1. Attachment to be uploaded
 /// 2. Attachment to be downloaded
 /// 3. Attachment to be deleted
-/// 3. Attachment to be archived
+/// 4. Attachment to be archived
 enum AttachmentState { queuedUpload, queuedDownload, queuedDelete, archived }
 
 class AttachmentsQueueTable extends Table {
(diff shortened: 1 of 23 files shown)


packages/powersync_attachments_helper/lib/src/attachments_queue_table.dart
```diff
@@ -64,11 +64,8 @@ class Attachment {
 /// 1. Attachment to be uploaded
 /// 2. Attachment to be downloaded
 /// 3. Attachment to be deleted
-enum AttachmentState {
-  queuedUpload,
-  queuedDownload,
-  queuedDelete,
-}
+/// 3. Attachment to be ignored
```
stevensJourney, line 67: Should this not be `4. archived`?

packages/powersync_attachments_helper/lib/src/attachments_queue_table.dart
```diff
@@ -64,11 +64,8 @@ class Attachment {
 /// 1. Attachment to be uploaded
 /// 2. Attachment to be downloaded
 /// 3. Attachment to be deleted
-enum AttachmentState {
-  queuedUpload,
-  queuedDownload,
-  queuedDelete,
-}
+/// 3. Attachment to be archived
```
stevensJourney, line 67:
```suggestion
/// 4. Attachment to be archived
```

stevensJourney (approved):

PR #68, opened 2024-03-31
fix: attachment queue duplicating requests

DominicGBauer:
## Description
Improve attachment queue so that multiple uploads, downloads and deletes don't run simultaneously. This is related to this support query in discord https://discord.com/channels/1138230179878154300/1220388800455250082

## Work Done 
* Replaced `watchDownloads`, `watchUploads` and `watchDeletes` into a single watcher `watchAttachments` so that duplicate upload, download or delete function calls do not occur when table is updated.
* Replaced `runDownloads`, `runUploads` and `runDeletes` into a single function `runSync` so that duplicate upload, download or delete function calls do not occur when table is updated.
* Updated dependencies
* Updated docs

## How to test
* Delete the demo app from the device you are using. Make sure you have set a bucket in config.
* Install the demo app and log in
* You should see a watcher run and begin downloading images. These should happen one at a time and not be duplicated.
* Now create a new todo and attach a photo.
* The upload should only occur once.

### Downloading
https://github.com/powersync-ja/powersync.dart/assets/46312751/273c52fa-d7a4-43cc-aec3-d6a11d3701f3

### Uploading
https://github.com/powersync-ja/powersync.dart/assets/46312751/06c75b1b-9b8d-4cfb-b4c5-568f34b9be45

### Deleting
https://github.com/powersync-ja/powersync.dart/assets/46312751/c94434f0-6d54-4618-82f1-1722c438edf6

### DIFF
--- a/demos/supabase-todolist/lib/attachments/queue.dart
+++ b/demos/supabase-todolist/lib/attachments/queue.dart
@@ -74,9 +74,9 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
       return results.map((row) => row['photo_id'] as String).toList();
     }).listen((ids) async {
       List<String> idsInQueue = await attachmentsService.getAttachmentIds();
-      for (String id in ids) {
-        await syncingService.reconcileId(id, idsInQueue, fileExtension);
-      }
+      List<String> relevantIds =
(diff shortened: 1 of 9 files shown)


stevensJourney (approved):
LGTM

PR #69, opened 2024-04-03
feat(powersync): add crud throttle

DominicGBauer:
## Description
Implement crud upload throttle as seen in `powersync-js/common-sdk` 

## Work Done
* Added throttle function 
* Cleaned up table names
* Did some refactoring to make things a bit neater
* Added crud throttle

## Todo
* I'm not sure if this is 100% complete

### DIFF
--- a/packages/powersync/lib/src/bucket_storage.dart
+++ b/packages/powersync/lib/src/bucket_storage.dart
@@ -6,6 +6,7 @@ import 'package:powersync/src/log_internal.dart';
 import 'package:sqlite_async/mutex.dart';
 import 'package:sqlite_async/sqlite3.dart' as sqlite;
 
+import 'powersync_database.dart';
 import 'crud.dart';
 import 'database_utils.dart';
 import 'schema_logic.dart';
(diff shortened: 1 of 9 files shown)


rkistner (commented):
Could you do the code style changes as a separate PR? It makes it much more difficult to find the actual changes in this PR.

I also don't like the table names being extracted into an enum - it makes the queries more verbose and more difficult to read, and breaks SQL language injection you get with IntelliJ IDEs.

Most of the other style changes are also just personal preferences, and I personally prefer the old style in most cases (e.g. if-else instead of switch or separate if statements).

rkistner (commented):
For the throttling logic:

From reading the code, I think the current implementation will miss updates in a case like this:
1. Write update A.
2. This triggers uploadCrud (not throttled).
3. Wait 10ms.
4. Write update B.
5. The update is throttled, does not trigger uploadCrud.

Now update B is not uploaded at all, until other update is written.

To fix that, the throttle implementation must make sure to also call uploadCrud on the trailing edge.


Regardless, it should not be needed to implement a new throttling function. There is already throttling implemented here:

https://github.com/powersync-ja/powersync.dart/blob/242636f0885c3edfbe190e4157226dfd07e2056f/packages/powersync/lib/src/powersync_database.dart#L215-L219

All that needs to be done is making that throttle duration configurable.

PR #78, opened 2024-04-23
feat(attachments)!: cater for subdirectories in storage

DominicGBauer:
## Description
* Allow subdirectories in storage to also be used in attachments queue
* Rename methods to make them more general

## Work Done
* Allow subdirectories to be set on queue initializing which will create a subdirectory locally to match what is on remote storage
* Renamed `savePhoto` and `deletePhoto` methods to `saveFile` and `deleteFile` - this introduces a BREAKING CHANGE

## How to test
Using todolist demo:
1. When initializing the queue in `queue.dart` add a subdirectory e.g. ['cover-images']
```
  PhotoAttachmentQueue(db, remoteStorage)
      : super(
            db: db,
            remoteStorage: remoteStorage,
            onDownloadError: onDownloadError,
            subdirectories: ['cover-images']);
```
2. Update `_takePhoto` method in `photo_capture_widget.dart` to this (bold is what has been changed, the ID now captures the subdirectory):
```
  Future<void> _takePhoto(context) async {
    try {
      // Ensure the camera is initialized before taking a photo
      await _initializeControllerFuture;

      final XFile photo = await _cameraController.takePicture();
      // copy photo to new directory with ID as name
      String subdirectory = 'cover-images'; <--------
      String photoId = '$subdirectory/${powersync.uuid.v4()}'; <--------
      String storageDirectory = await attachmentQueue.getStorageDirectory();
      await attachmentQueue.localStorage
          .copyFile(photo.path, '$storageDirectory/$photoId.jpg');

      int photoSize = await photo.length();

      TodoItem.addPhoto(photoId, widget.todoId);
      attachmentQueue.saveFile(photoId, photoSize);
    } catch (e) {
      log.info('Error taking photo: $e');
    }

    // After taking the photo, navigate back to the previous screen
    Navigator.pop(context);
  }
  ```
3. Load the app and take a photo after creating a todo
4. You should see the image in a subfolder on supabase
5. Load the app on another device and you should see the image download onto the device and display correctly

### DIFF
--- a/demos/supabase-todolist/lib/attachments/queue.dart
+++ b/demos/supabase-todolist/lib/attachments/queue.dart
@@ -39,10 +39,11 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
   }
 
   @override
-  Future<Attachment> savePhoto(String photoId, int size) async {
-    String filename = '$photoId.jpg';
+  Future<Attachment> saveFile(String fileId, int size) async {
+    String filename = '$fileId.jpg';
+
     Attachment photoAttachment = Attachment(
-      id: photoId,
+      id: fileId,
       filename: filename,
       state: AttachmentState.queuedUpload.index,
       mediaType: 'image/jpeg',
@@ -54,10 +55,11 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
   }
 
   @override
-  Future<Attachment> deletePhoto(String photoId) async {
-    String filename = '$photoId.jpg';
+  Future<Attachment> deleteFile(String fileId) async {
+    String filename = '$fileId.jpg';
+
     Attachment photoAttachment = Attachment(
-        id: photoId,
+        id: fileId,
         filename: filename,
         state: AttachmentState.queuedDelete.index);
 
--- a/demos/supabase-todolist/lib/attachments/queue.dart
+++ b/demos/supabase-todolist/lib/attachments/queue.dart
@@ -39,14 +39,15 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
   }
 
   @override
-  Future<Attachment> saveFile(String fileId, int size) async {
+  Future<Attachment> saveFile(String fileId, int size,
+      {mediaType = 'image/jpeg'}) async {
     String filename = '$fileId.jpg';
 
     Attachment photoAttachment = Attachment(
       id: fileId,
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: demos/supabase-todolist/lib/attachments/queue.dart [hunk 1]
```diff
@@ -39,10 +39,11 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
   }
 
   @override
-  Future<Attachment> savePhoto(String photoId, int size) async {
-    String filename = '$photoId.jpg';
+  Future<Attachment> saveFile(String fileId, int size) async {
+    String filename = '$fileId.jpg';
```
mugikhan, line 43: This still seems like it would be for photo uploads. Should the `extension` and `mediaType` not be configurable or passed in? It could be defaulted to an image.

DominicGBauer: Yeah nice catch!

demos/supabase-todolist/lib/attachments/queue.dart
```diff
@@ -54,10 +56,11 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
   }
 
   @override
-  Future<Attachment> deletePhoto(String photoId) async {
```
stevensJourney, line 57: Do we need the breaking change of renaming `deletePhoto` etc? Could we not preserve this method and call the generic `xxxFile` method under the hood?

DominicGBauer: Since this is still in Beta I would rather remove it to avoid confusion for new devs coming in. I think we have at most  a handful of users using this (based on support queries). But I don't feel strongly about this.

stevensJourney: I'm hesitant to include breaking changes in minor package version bumps. This could cause headaches for existing users.

DominicGBauer: Isn't that the point of version 0 though so people know if there's a minor upgrade it could have breaking changes so should rather use ~ instead of ^?

stevensJourney: Yeah, you're right. I forgot about the magic of version `0`.

Chriztiaan (approved):

PR #81, opened 2024-04-29
Reset isProcessing when exception is thrown during sync process.

ScottAtRedHawk:
The `isProcessing` property needs to be reset if an error is thrown during a sync so that subsequent sync attempts can be made. If it's not reset, then it's always `true` and will never sync again until the app is closed and re-opened.


DominicGBauer (approved):
Thanks for catching this and making a PR @ScottAtRedHawk!

PR #95, opened 2024-06-13
Upgrade powersync-sqlite-core to version 0.1.7 for Android

mugikhan:
## Work done

The work was verified by running it on an android emulator (API 21), ensuring the extension loads and `db.get('select powersync_rs_version()')` returns version 0.1.7.

- Upgrade powersync-sqlite-core to version 0.1.7 for Android
- Upgrade `powersync_flutter_libs` to version 0.1.0
- Bump `powersync` to version 1.4.1


mugikhan:
I've also updated all the versions of the demos powersync packages to the latest. This is so that if someone copies or tries to run the demo outside the mono repo is still works.

rkistner:

PR #99, opened 2024-06-18
Override default macOS sqlite db for tests

mugikhan:
## Description 

Override the default macOS SQLite binary when running tests as it does not allow extension loading. This allows the extension to be loaded when running tests on macOS.

This requires the sqlite binary and extension binary to be placed in the root folder for testing.

## Work done

- Override the default macOS SQLite loaded for tests.
- Load the dynamic library for all platforms.

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -31,7 +31,7 @@ scripts:
 
   test:
     description: Run tests in a specific package.
-    run: dart test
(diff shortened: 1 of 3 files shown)


rkistner (approved):

rkistner:

PR #101, opened 2024-06-20
Add recursive triggers for web

mugikhan:


packages/powersync/lib/src/web/powersync_db.worker.dart
```diff
@@ -21,6 +21,7 @@ final class PowerSyncAsyncSqliteController extends AsyncSqliteController {
   Future<WorkerDatabase> openDatabase(
       WasmSqlite3 sqlite3, String path, String vfs) async {
     final db = sqlite3.open(path, vfs: vfs);
+    db.execute('PRAGMA recursive_triggers = TRUE');
```
stevensJourney, line 24: This pragma statement is already present [here](https://github.com/powersync-ja/powersync.dart/blob/171f848a983ee304b362ca307235127b1cd7db2e/packages/powersync/lib/src/open_factory/abstract_powersync_open_factory.dart#L26)

Seems like the issue is that the `open` method of `AbstractPowerSyncOpenFactory` executes these pragma statements. The `open` method is used by the `openConnection` method for native implementations, but is not used for web. 

I think we should be able to override the `openConnection` method in [here](https://github.com/powersync-ja/powersync.dart/blob/sqlite_proto/packages/powersync/lib/src/open_factory/web/web_open_factory.dart) To execute relevant pragmas.

packages/powersync/lib/src/open_factory/web/web_open_factory.dart
```diff
@@ -14,6 +17,16 @@ class PowerSyncOpenFactory extends AbstractPowerSyncOpenFactory {
     // No op for web
   }
 
+  @override
+  Future<SqliteConnection> openConnection(SqliteOpenOptions options) async {
+    var conn = await super.openConnection(options);
+    for (final statement in super.pragmaStatements(options)) {
+      conn.execute(statement);
```
stevensJourney, line 24: should we await this?

mugikhan: Yes, `execute` is a future. Fixed it.

stevensJourney (approved):

rkistner:

PR #105, opened 2024-06-25
Upgrade sqlite_async

mugikhan:
## Description

This upgrades `sqlite_async` to version 0.7.0 which requires the minimum dart SDK version 3.4.0 and fixes the `Invalid argument(s): argument value for 'return_value' is null` error when closing the database connection.
This also updates all the Database types to a `CommonDatabase` type.


stevensJourney (approved):

rkistner:
The tests failed because of a flaky test, safe to ignore (fixed in later versions).

PR #108, opened 2024-06-28
Django Todolist Demo App

kobiebotha:
Pairs with https://github.com/powersync-ja/powersync-django-backend-todolist-demo/pull/3


demos/django-todolist/lib/apiclient.dart
```diff
@@ -0,0 +1,60 @@
+import 'dart:convert';
```
DominicGBauer, line 1: Can you use snake case for this file name i.e. api_client to conform to files names in the repo

PR #113, opened 2024-07-02
chore(attachments): change uploadfile and deletefile return types

DominicGBauer:
## Description
Allow users to return what they would like from `uploadFile` and `deleteFile` as we currently restrict them to `void` which has become an issue for certain users.

## Work Done
* Updated return type to `dynamic`
* Updated demo


stevensJourney:
Just for my context: What is the use case for the returned value from `uploadFile` etc? I thought this method is called internally by the attachment queue?

DominicGBauer:
@stevensJourney a user wants to use the URL returned from supabase when uploading

DominicGBauer:
Closed this as this doesn't actually solve a problem

PR #115, opened 2024-07-02
Close http connections immediately on disconnect

### LINKED ISSUE #114: HTTP connections not immediately closed on `db.disconnect()`
When calling `db.disconnect()`, the sync Isolate is closed, but the HTTP connection still remains open for a while.

This does not cause any consistency issues on the client, but this counts towards concurrent connections and operations synced on the server.

This is especially noticeable when connecting and disconnecting in a loop, which can cause concurrent connections on the server to keep on increasing for a couple of minutes.

rkistner:
Fixes #114.

This now ensures the http connection is properly closed when disconnecting, before killing the Isolate.

This waits for any requests in progress, since closing the http client while the request is starting causes unpredictable uncaught errors.

It is tricky to cover all cases, but the tests do a fairly good job of checking them. This PR also improves the connection counting in tests, avoiding some sporadic test failures we had previously.

### DIFF
--- a/packages/powersync/lib/src/powersync_database.dart
+++ b/packages/powersync/lib/src/powersync_database.dart
@@ -191,13 +191,24 @@ class PowerSyncDatabase with SqliteQueries implements SqliteConnection {
       /// Throttle time between CRUD operations
       /// Defaults to 10 milliseconds.
       Duration crudThrottleTime = const Duration(milliseconds: 10)}) async {
-    _connectMutex.lock(() =>
-        _connect(connector: connector, crudThrottleTime: crudThrottleTime));
+    Zone current = Zone.current;
+
+    Future<void> reconnect() {
+      return _connectMutex.lock(() => _connect(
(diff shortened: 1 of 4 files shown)


packages/powersync/lib/src/powersync_database.dart
```diff
@@ -625,6 +642,7 @@ Future<void> _powerSyncDatabaseIsolate(
         uploadCrud: uploadCrud,
         updateStream: updateController.stream,
         retryDelay: args.retryDelay);
+    openedStreamingSync = sync;
```
mugikhan, line 645: Small question on this. Why not just assign `StreamingSyncImplementation` on line 638 to `openedStreamingSync`?

rkistner: Mostly because it would then need null guards in all other places, since `openedStreamingSync` is nullable.

mugikhan (approved):

PR #117, opened 2024-07-03
Fix Future already completed when calling waitForFirstSync

mugikhan:
## Work done

- Refactor `waitForFirstSync()` to iterate through the stream and remove the use of a `Future`.

### DIFF
--- a/packages/powersync/CHANGELOG.md
+++ b/packages/powersync/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.5.2
+
+- Fix `Bad state: Future already completed` error when calling `waitForFirstSync()`.
+
 ## 1.5.1
 
 - Adds a hasSynced flag to check if initial data has been synced.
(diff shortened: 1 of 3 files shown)


rkistner (commented):
Looking at this now, the statusStream subscription should also be cancelled when completing the future.

An alternative that may be simpler is to iterate through the stream, avoiding the need for a separate completer or managing the subscription:

```dart
await for (final result in stream) {
  if (result.hasSynced ?? false) {
    break;
  }
}
```

mugikhan:
> Looking at this now, the statusStream subscription should also be cancelled when completing the future.
> 
> An alternative that may be simpler is to iterate through the stream, avoiding the need for a separate completer or managing the subscription:
> 
> ```dart
> await for (final result in statusStream) {
>   if (result.hasSynced ?? false) {
>     break;
>   }
> }
> ```

This is a more elegant solution and it works as expected. Thanks

rkistner (dismissed):

rkistner (approved):

PR #118, opened 2024-07-04
[Chore] Update Sqlite Async Web

stevensJourney:
# Overview

This updates the `sqlite_async` package to `0.8.0` which now contains web support.

Testing: Verified that the Supabase Todolist app works as expected.

### DIFF
--- a/demos/supabase-todolist/pubspec.yaml
+++ b/demos/supabase-todolist/pubspec.yaml
@@ -10,9 +10,9 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  powersync_attachments_helper: ^0.3.0-alpha.2
+  powersync_attachments_helper: ^0.3.0-alpha.3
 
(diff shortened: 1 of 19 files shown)


demos/supabase-todolist/pubspec.yaml
mugikhan, line 1: I think we should also update all the demos to pin the version of powersync to `powersync: 1.3.0-alpha.7`

mugikhan (changes requested):

mugikhan (approved):

demos/supabase-todolist/pubspec.yaml
stevensJourney: Fair enough. I've pinned those now. For context the `powersync_attachments_helper`'s version was pinned to avoid the Pana package downgrade test from failing, but it can't hurt to pin the demos.

rkistner:

Issue #11, opened 2023-08-13
Web support?


matthewwong525:
Web is a key platform for my application and it's the main blocker for me switching to powersync right now! Thanks for the work team :)

cahofmeyr:
@matthewwong525 Thanks for the feedback! Web support is on the way. We will post updates here once we have more specifics to share :)

cahofmeyr:
@matthewwong525 quick question: Do you need support for Flutter for web, or for non-Flutter web apps?

Also, we'd love to talk to you for 5-10mins to hear more about your requirements and ask a couple of other questions. If you're open to this, could you drop me a line at conrad[at]powersync.co? Thanks!

matthewwong525:
> @matthewwong525 quick question: Do you need support for Flutter for web, or for non-Flutter web apps?
> 
> Also, we'd love to talk to you for 5-10mins to hear more about your requirements and ask a couple of other questions. If you're open to this, could you drop me a line at conrad[at]powersync.co? Thanks!

My app uses flutter so I only need Flutter for web support. I don't think non-flutter web apps are part of it. For some context, I'm building a note-taking app that supports all platforms through Flutter.

cahofmeyr:
@matthewwong525 thanks!

kamranbekirovyz:
Waiting updates on WEB support to plan using powersync 🫠

kobiebotha:
> Waiting updates on WEB support to plan using powersync 🫠

Sit tight, we're working on it! 

What's your web stack btw?

benitav:
Hey everyone! I am pleased to share that we now have web support in an alpha release. It is open for anyone to try: https://www.powersync.com/blog/announcing-powersync-flutter-web-support

oleksiiparf:
Hey guys! Please take into account that with the current alpha version **PowerSyncDatabase.disconnectAndClear()** functionality is not working as expected - during logout the data in tables is not deleted at all and can be accessed with different user session, which is a major issue.

Steps to reproduce:
1. Login as User A to Supabase. Data(lists, todos) can be pre-created for illustration.
2. Logout from Supabase. Call PowerSyncDatabase.disconnectAndClear().
3. Login as User B to Supabase. Before the initial data will be downloaded for User B I'll see the data from previous session of User A.

benitav:
@oleksiiparf thanks for reporting! We can confirm the issue and are working on a fix.

stevensJourney:
[...]

PR #119, opened 2024-07-09
[Feature] Implement web sync locks

stevensJourney:
# Overview

This uses the Navigator locks in the `Mutex` class from here https://github.com/powersync-ja/sqlite_async.dart/pull/54 to limit the number of active sync implementations when multiple tabs are used.

This is only a stop-gap to prevent multiple tabs with multiple `StreamingSyncImplementation`s from applying sync bucket changes at the same time (this can leave the local SQLite DB in an inconsistent or broken state).

This method enhances the safety of using multiple tabs, but is far from complete multiple tab support. 

This should address the issue of "Flutter web + PowerSync still has the issue where multiple open tabs will cause sqlite to get messed up"  from https://github.com/powersync-ja/powersync.dart/issues/11

## Testing

Note that the app works when switching between tabs. The sync status is isolated per tab - only 1 tab is ever connected or uploading at a time. 

https://github.com/powersync-ja/powersync.dart/assets/51082125/a5eab365-45d9-4523-b485-e110ea509e3e

## Todos:
- [x] Update sqlite_async package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,34 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2024-07-10
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+ - There are no breaking changes in this release.
+
+Packages with other changes:
+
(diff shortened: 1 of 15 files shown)


rkistner (commented):
The approach here looks good to me

mugikhan (approved):

PR #98, opened 2024-06-14
[Feature] Added user parameters support when connecting 

Chriztiaan:
This PR aims to introduce the same functionality introduced in `Powersync-js` [here](https://github.com/powersync-ja/powersync-js/pull/204). It introduces the ability to pass parameters from a client SDK to the sync rules. The idea is to specify parameters to the PowerSync `connect` method's options. Backend changes to support this are available [here](https://github.com/powersync-ja/powersync-service/pull/7). 

```
// demos/supabase-todolist/lib/powersync.dart
db.connect(connector: currentConnector, params: {"device_id": "6e555518-6ed5-4ca4-b5b7-e1624b995aca"});
  ```

Parameters are available in the sync rules under `user_parameters` (alongside the already supported `token_parameters`).

For example:
```
bucket_definitions:
  user_lists:
    # Separate bucket per todo list
    parameters: select id as list_id from lists where owner_id = user_parameters.device_id
    data:
      - select * from lists where id = bucket.list_id
      - select * from todos where list_id = bucket.list_id
````

PR #122, opened 2024-07-11
Merge master into alpha release

mugikhan:


.github/workflows/demos.yml
```diff
@@ -38,11 +38,16 @@ jobs:
...
+          channel: "stable"
       - name: Install melos
         run: flutter pub global activate melos
       - name: Install dependencies
         run: melos prepare
+      - name: Download powersync binary
+        run: |
+          github="https://github.com/powersync-ja/powersync-sqlite-core/releases/download/v0.1.6"
+
+          curl "${github}/libpowersync_x64.so" -o packages/powersync/libpowersync.so --create-dirs -L -f
```
stevensJourney, line 51: I feel like we should have a Dart Script which downloads the platform specific binary - to make running tests locally easier for developers. 

We currently have scripts for web workers and SQLite WASM which are triggered when running `melos prepare` - I think it should be relatively easy to add a script for this.

melos.yaml
```diff
@@ -46,7 +46,7 @@ scripts:
 
   test:
     description: Run tests in a specific package.
-    run: dart test -p vm,chrome
+    run: flutter test
```
stevensJourney, line 49: Will this still test `chrome`?

.github/workflows/demos.yml
mugikhan: The only problem is that this needs to be platform specific. So the script will have to detect their platform and download the respective binary.

stevensJourney: The platform should be available in the Dart script with:

```Dart
import 'dart:io' show Platform;
```
which is the reason for recommending using a Dart script.

packages/powersync/lib/src/bucket_storage.dart
```diff
@@ -253,7 +104,7 @@ class BucketStorage {
 
     if (!r.checkpointValid) {
       for (String b in r.checkpointFailures ?? []) {
-        await deleteBucket(b);
+        deleteBucket(b);
```
stevensJourney, line 107: Don't we need to await this still?

packages/powersync/lib/src/bucket_storage.dart
```diff
@@ -662,14 +289,13 @@ class BucketStorage {
         haveMore: true,
         complete: ({String? writeCheckpoint}) async {
           await writeTransaction((tx) async {
-            await tx
-                .execute('DELETE FROM ps_crud WHERE id <= ?', [last.clientId]);
+            tx.execute('DELETE FROM ps_crud WHERE id <= ?', [last.clientId]);
```
stevensJourney, line 292: Seems like these are missing `await`

packages/powersync/lib/src/database/native/native_powersync_database.dart
```diff
@@ -54,6 +54,8 @@ class PowerSyncDatabaseImpl
   /// Use [attachedLogger] to propagate logs to [Logger.root] for custom logging.
   late final Logger logger;
 
+  Map<String, dynamic>? clientParams;
```
stevensJourney, line 57: I'm not sure why this is public or what it's used for - but I think this should be on `PowerSyncDatabaseMixin` in order for it to be available on all the implementations.

packages/powersync/lib/src/database/native/native_powersync_database.dart
```diff
@@ -130,7 +132,10 @@ class PowerSyncDatabaseImpl
 
       /// Throttle time between CRUD operations
       /// Defaults to 10 milliseconds.
-      Duration crudThrottleTime = const Duration(milliseconds: 10)}) async {
+      required Duration crudThrottleTime,
+      required Future<void> Function() reconnect,
+      Map<String, dynamic>? params}) async {
+    clientParams = params;
```
stevensJourney, line 138: This could be done in the connect method of `PowerSyncDatabaseMixin`

packages/powersync/lib/src/database/web/web_powersync_database.dart
```diff
@@ -144,13 +145,14 @@ class PowerSyncDatabaseImpl
...
         client: FetchClient(mode: RequestMode.cors),
+        syncParameters: params,
         // Only allows 1 sync implementation to run at a time per database
         // This should be global (across tabs) when using Navigator locks.
         identifier: database.openFactory.path);
     sync.statusStream.listen((event) {
       setStatus(event);
     });
-    sync.streamingSync(abortController: disconnecter);
+    sync.streamingSync();
```
stevensJourney, line 155: The `disconnecter` is used when the `disconnect` method is called, but now this is no longer tied to the `StreamingSyncImplementation`. I think we should be able to do (taken from Native implementation):

```Dart
 disconnector.onAbort.then((_) {
           sync.abort();
          }).ignore();
```

packages/powersync/lib/src/database/web/web_powersync_database.dart
```diff
@@ -189,22 +191,22 @@ class PowerSyncDatabaseImpl
...
-  Future<T> writeTransaction<T>(
-      Future<T> Function(SqliteWriteContext tx) callback,
-      {Duration? lockTimeout,
-      String? debugContext}) async {
-    await isInitialized;
-    return database.writeTransaction(
-        (context) => internalTrackedWrite(context, callback),
-        lockTimeout: lockTimeout);
-  }
+  // Future<T> writeTransaction<T>(
```
stevensJourney, line 194: The `writeTransaction` method on the web PowerSync database client had an override to avoid using the default implementation provided by the `SqliteQueries` [mixin](https://github.com/powersync-ja/sqlite_async.dart/blob/9428ad9ec49f0a12a5c0bf351b6bc56b359f99bb/packages/sqlite_async/lib/src/sqlite_queries.dart#L111) from `sqlite_async`. If the mixin was used it would start a `writeLock` on the `database` and then manually start a transaction.

The reason for the override is that the underlaying web `database` has custom functions for transaction SQL execution in it's `writeTransaction` method. We want to use that logic on the PowerSyncDatabase level - so we avoid the mixin implementation and call the underlaying `database` writeTransaction method instread.

packages/powersync/lib/src/stream_utils.dart
```diff
@@ -68,6 +68,13 @@ Stream<Object?> ndjson(ByteStream input) {
   return jsonInput;
 }
 
+/// Given a raw ByteStream, parse each line as JSON.
+Stream<String> newlines(ByteStream input) {
```
stevensJourney, line 72: I think this isn't actually used anywhere.

packages/powersync/lib/src/database/powersync_db_mixin.dart
```diff
@@ -78,7 +75,6 @@ mixin PowerSyncDatabaseMixin implements SqliteConnection {
         .cast<UpdateNotification>();
 
     await database.initialize();
-    await migrations.migrate(database);
     await updateSchema(schema);
```
stevensJourney, line 78: The `hasSynced` init should be added here https://github.com/powersync-ja/powersync.dart/blob/9e1e376a1b1c18c8347a3c2a3bd056731296f305/packages/powersync/lib/src/powersync_database.dart#L158

.github/workflows/demos.yml
mugikhan: Done, had to check the Abi specifically because of the differences between x64 and arm.

packages/powersync/lib/src/bucket_storage.dart
```diff
@@ -58,155 +40,33 @@ class BucketStorage {
...
     await writeTransaction((tx) async {
       for (var b in batch.buckets) {
-        var bucket = b.bucket;
-        var data = b.data;
-
-        count += data.length;
-        final isFinal = !b.hasMore;
-        await _updateBucket(tx, bucket, data, isFinal);
+        count += b.data.length;
+        await _updateBucket2(
```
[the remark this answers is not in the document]
mugikhan: I think this is what Ralf named it in the original refactor and is what currently exists on master. Should I rename it?

packages/powersync/lib/src/database/web/web_powersync_database.dart
mugikhan: The concern I have with this is with the new custom sqlite wasm that has the rust methods, will it not execute `UPDATE ps_tx SET current_tx = next_tx, next_tx = next_tx + 1 WHERE id = 1` in  the `internalTrackedWrite()` function?

packages/powersync/lib/src/stream_utils.dart
mugikhan: Removed

packages/powersync/lib/src/database/native/native_powersync_database.dart
mugikhan: Its related to this for adding user parameters support (#98).

packages/powersync/lib/src/database/web/web_powersync_database.dart
mugikhan: Done! Thanks for the suggestion.

packages/powersync/lib/src/database/web/web_powersync_database.dart
stevensJourney: My understanding is that the rust core intercepts [begin](https://github.com/powersync-ja/powersync-sqlite-core/blob/5704c7fb689b84f8679b006023f3fd0858fd75ff/crates/core/src/crud_vtab.rs#L171) events and automatically tracks those transactions. I don't think we are creating the transaction multiple times - so I don't think anything should be tracked twice.

stevensJourney:
Tested this on most platforms - things seem to work normally.

An observation in web: due to the sync status not being shared in multiple tabs the `hasSynced` helper does not currently work in multiple tabs.


Uploading testalpha.mp4…

mugikhan:
> Tested this on most platforms - things seem to work normally.
> 
> An observation in web: due to the sync status not being shared in multiple tabs the `hasSynced` helper does not currently work in multiple tabs.
> 
>  testalpha.mp4

Do we want to note this as a known issue or should I look into fixing it for alpha?

packages/powersync/lib/src/database/web/web_powersync_database.dart
rkistner: That is correct - the rust core now handles that (and only increments the transaction number when actually writing to ps_crud).

stevensJourney:
> > Tested this on most platforms - things seem to work normally.
> > An observation in web: due to the sync status not being shared in multiple tabs the `hasSynced` helper does not currently work in multiple tabs.
> > testalpha.mp4
> 
> Do we want to note this as a known issue or should I look into fixing it for alpha?

This issue seems to affect `master` as well if the device is offline. I created a PR https://github.com/powersync-ja/powersync.dart/pull/124 for the fix. We can update it here (there will be conflicts) after merging.

packages/powersync/lib/src/bucket_storage.dart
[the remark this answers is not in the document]
stevensJourney: Having the `2` there without another version seems weird to me. I can't think of why it needs to be that way. But if it's in master like that then we can update that later.

melos.yaml
stevensJourney: I don't see chrome tests here https://github.com/powersync-ja/powersync.dart/actions/runs/9944203557/job/27469732966

mugikhan: @stevensJourney Yeah the tests fail when its run with `dart -p chrome`. I will look into it again as the native side needs `flutter test` for loading the native extension files.

packages/powersync/lib/src/bucket_storage.dart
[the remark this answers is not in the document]
rkistner: It was there when I did the initial migration - `_updateBucket` was the version with plain Dart code, `_updateBucket2` was the one using powersync-sqlite-core. I'm happy for it to just be `_updateBucket` again, and don't have a strong opinion on when we rename it.

stevensJourney (approved):
LGTM

PR #124, opened 2024-07-16
[Fix] Offline has synced

stevensJourney:
# Overview

Currently the `hasSynced` member of `SyncStatus` does not work if the device is offline.

The value is initially set to true after querying `ps_buckets` but if a sync request fails it's value will be cleared.

# Testing

## Before

When starting offline
image

Note that `lists` are present on the device
image

## After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,32 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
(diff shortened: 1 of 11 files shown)


mugikhan (dismissed):
LGTM.

mugikhan (approved):

PR #126, opened 2024-07-22
Custom dart script to download wasm and worker files

mugikhan:
## Description

Introduces a custom script designed to download WebAssembly (WASM) and worker files. The script automates the process of retrieving these files by running a single command `dart run powersync:setup_web` in the application folder.

## Work done

- Create a custom dart script to download `sqlite3.wasm` and `powersync_db.worker.js` files.

## TODO

- [x] Fix versioning for `sqlite3.wasm` releases so that it can be parsed from a pubspec file. PR [here](https://github.com/powersync-ja/sqlite3.dart/pull/2)


packages/powersync/bin/setup_web.dart
```diff
@@ -0,0 +1,96 @@
...
+  final sqlite3Version = parsed.version?.toString();
+  if (sqlite3Version == null) {
+    print('Sqlite3 version not found');
+    print('Run `flutter pub get` first.');
+    exit(1);
+  }
+
+  //TODO: Use `sqlite3Version` to get the correct sqlite3.wasm
+  final sqliteUrl =
+      'https://github.com/powersync-ja/sqlite3.dart/releases/download/v0.1.0/sqlite3.wasm';
```
rkistner, line 63: This bit can be tricky - I believe we actually need a wasm build with _both_ the correct sqlite3 version and powersync-sqlite-core version?

mugikhan: @rkistner The idea is to have our custom builds to use the versions from the `sqlite3` pubspec. That way we can always fetch the latest build and release a new build (update the fork) every time the upstream repository is updated. Keeping the core in sync with the sqlite build will be a bit tricky right now and will have to be done manually for now.

packages/powersync/README.md
```diff
@@ -49,13 +49,11 @@ The latest prerelease version can be found [here](https://pub.dev/packages/power
 
 ### Additional config
 
-Web support requires `sqlite3.wasm` and `powersync_db.worker.js` assets to be served from the web application. This is typically achieved by placing the files in the project `web` directory.
+Web support requires `sqlite3.wasm` and `powersync_db.worker.js` assets to be served from the web application. They can be downloaded to the `web` directory by running the following command in your applications root folder.
```
stevensJourney, line 52:
```suggestion
Web support requires `sqlite3.wasm` and `powersync_db.worker.js` assets to be served from the web application. They can be downloaded to the `web` directory by running the following command in your application's root folder.
```

packages/powersync/bin/setup_web.dart
```diff
@@ -0,0 +1,129 @@
...
+
+  final sqlitePackageName = 'sqlite3';
+
+  final sqlite3Pkg = getPackageFromConfig(packageConfig, sqlitePackageName);
+
+  String sqlite3Version =
+      "v${getPubspecVersion(packageConfigFile, sqlite3Pkg, sqlitePackageName)}";
+  final httpClient = HttpClient();
+
+  String latestTag = await getLatestTagFromRelease(httpClient);
```
stevensJourney, line 53: The logic here doesn't seem complete. 

It looks like it's getting the latest release tag from all the releases and then using it if the tag starts with the `sqlite3Version`. 

This might be okay for some scenarios, but it will not give the latest `-powersync.x` version if the `sqlite3Version` in use in the application does not match our latest release. For example if the application is using `sqlite3:2.3.0` and our latest release is `2.4.0-powersync.1` but we also had `2.3.0-powersync.4` then it will only use `2.3.0-powersync.1`

scripts/init_sqlite_wasm.dart
```diff
@@ -1,8 +1,8 @@
-/// Downloads sqlite3.wasm and copies it to all demo folders
+/// Downloads sqlite3.wasm into the assets directory for tests
```
stevensJourney, line 1: Do we still need this script? Can't the new tool download this for us?

melos.yaml
```diff
@@ -9,7 +9,19 @@ ide:
   intellij: false
 
 scripts:
-  prepare: melos bootstrap && dart ./scripts/compile_webworker.dart && dart ./scripts/init_sqlite_wasm.dart && dart ./scripts/init_powersync_core_binary.dart
+  prepare:
+    description: Download SQLite3 WASM for demos
+    run: |
+      melos bootstrap && dart ./scripts/init_sqlite_wasm.dart && dart ./scripts/compile_webworker.dart && dart ./scripts/init_powersync_core_binary.dart
+      for dir2 in demos/*; do
```
stevensJourney, line 16: Will this will work on all operating systems? I think we've had some reported issues of people using this repo on Windows before where some shell scripts did not work. 

Is it not possible to use Melos to execute the `dart run powersync:setup_web no_worker` command on the filtered monorepo folders?

scripts/init_sqlite_wasm.dart
mugikhan: The problem with using the new script is that it is run multiple times for the demos and if we add it to download to the assets it will download multiple times for each demo that has a web folder. If you look at the `melos prepare` command thats how it currently functions. I could rework it to add an additional argument that downloads it to assets just once.

The other consideration is that users will be running the script and they don't need that functionality for their applications.

melos.yaml
mugikhan: As far as I understand it with `melos`, we can filter multiple package directories to run a command in each folder. I don't think it is possible to filter arbitrary directories like the `demos` directories and run a command in each through `melos`

packages/powersync/bin/setup_web.dart
mugikhan: I reworked it to match a tag from the github releases, if nothing is matched it uses the latest version.

melos.yaml
mugikhan: Reworked it to run with `melos`.

scripts/init_sqlite_wasm.dart
mugikhan: File has been removed in favor of an output directory argument.

PR #136, opened 2024-08-01
Automate package publishing

mugikhan:
## Description

When trying to build the project for windows or linux the user is required to have an internet connection in order to download the respective powersync-core binaries. This change bundles the binaries into the `powersync_flutter_libs` package when publishing the package and allows the build to be completed irrespective of the network connection.

## Work done

- Update publish packages workflow
- Remove download and update powersync core binary filenames for Linux and Windows

## TODO: 

- [x] Enable windows arm64 builds for powersync-core (Postponed)


RELEASING.md
```diff
@@ -0,0 +1,26 @@
...
+This will create a tag in the format of ${PACKAGE_NAME}-vm.m.p
+
+```
+e.g powersync-v1.6.4, powersync_attachments_helper-v0.6.3+1, etc.
+```
+
+# Perform Release
+
+```
+git push --follow-tags
```
stevensJourney, line 25: Can we also mention about the Draft Github Release which is created? That currently needs to be manually checked and published after it's created.

mugikhan: Added a note at the end to publish the draft release.

PR #139, opened 2024-08-05
[Fix]  Reset hasSynced

stevensJourney:
# Overview

The `hasSycned` status would be retained after calls to `disconnectAndClear` have been made. 

Steps to reproduce the issue:

- Log in
- Wait for first sync > works fine
- Log out
- Log in
- Wait for first sync > no longer works

The previous `hasSynced: true` would cause the check to immediately pass before syncing had completed on the second round. 

This PR clears the `hasSynced` status when disconnecting and clearing. The PowerSync client typically checks the setting in the `initialize` method which won't be called again, but the status will be updated on the next full sync when `lastSyncedAt` has been set.

# Testing

This was tested on MacOS by syncing locally then logging out. The local PowerSync service was then stopped to delay the syncing process. The next login should show the loading message until a sync has occurred.

The first test is without the fix, the second test contains the fix.

https://github.com/user-attachments/assets/47be618f-a883-47fc-82cc-13bf51d5ad2e

On web it was observed that a large queue of sync status updates were present after `disconnectAndClear` this was due to the `crudLoop` not ending when the `StreamingSyncImplementation` was aborted. A break for uploads was added.

This only shows the fix present.

https://github.com/user-attachments/assets/8f9520e7-a2f5-47f6-bdd4-85e71c4d1dd0

### DIFF
--- a/packages/powersync/lib/src/database/powersync_db_mixin.dart
+++ b/packages/powersync/lib/src/database/powersync_db_mixin.dart
@@ -239,6 +239,8 @@ mixin PowerSyncDatabaseMixin implements SqliteConnection {
         await tx.execute('DELETE FROM ${quoteIdentifier(row['name'])}');
       }
     });
+    // The data has been deleted - reset these
+    setStatus(SyncStatus(lastSyncedAt: null, hasSynced: false));
   }
 
   @Deprecated('Use [disconnectAndClear] instead.')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ Packages with dependency updates only:
 
 #### `powersync` - `v1.6.4`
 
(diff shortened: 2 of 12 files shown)


mugikhan (approved):

packages/powersync/lib/src/database/powersync_db_mixin.dart
```diff
@@ -239,6 +239,8 @@ mixin PowerSyncDatabaseMixin implements SqliteConnection {
         await tx.execute('DELETE FROM ${quoteIdentifier(row['name'])}');
       }
     });
+    // The data has been deleted - reset these
+    setStatus(SyncStatus(lastSyncedAt: null, hasSynced: false));
```
mugikhan, line 243: Do we have this same issue on the JS SDK? I see we do not reset the status in `disconnectAndClear()`.

stevensJourney: We might have the same issue there also. I haven't tested it yet, but from looking at the code quickly it seems like it should have the same issue.

mugikhan (approved):

PR #148, opened 2024-08-20
powersync-sqlite-core v0.2.1

rkistner:
Update with the changes in https://github.com/powersync-ja/powersync-sqlite-core/pull/21.

When using a self-hosted instance, this requires v0.5.0 or later of the image, to include [these changes](https://github.com/powersync-ja/powersync-service/pull/64). The changes are already rolled out on all hosted instances.

## Changes

1. Customize User-Agent header. Before: `Dart/3.4 (dart:io)`. Now: `powersync-dart/1.6.4 (Dart/3.4.2; linux)`
2. Add `client_id` parameter with an unique uuid per client.
3. Persist `lastSyncedAt`.
4. Emit update notifications on `disconnectAndClear()`.
5. Validate that the powersync-sqlite-core version number is in a compatible range.
6. Always cast `target_op` to ensure it's an integer.
7. Sync optimizations: https://github.com/powersync-ja/powersync-sqlite-core/pull/19


stevensJourney (approved):
LGTM

PR #149, opened 2024-08-21
Bump versions

mugikhan:
## Work done

- Bump versions for automated publishing
- Fixes adding downloaded binaries to `powersync_flutter_libs` when publishing


.github/workflows/publish.yml
```diff
@@ -39,6 +39,10 @@ jobs:
           echo "PACKAGE_NAME=$PACKAGE_NAME" >> $GITHUB_ENV
           echo "Package name: $PACKAGE_NAME"
 
+      - name: Update gitignore
+        run: |
+          sed -i '' -e 's/*.so/!*.so/g; s/*.dll/!*.dll/g; s/*.dylib/!*.dylib/g' packages/powersync_flutter_libs/.gitignore
+
```
rkistner: Instead of updating the .gitignore file, you can create a `.pubignore` file that takes precedence over `.gitignore` for publishing.

https://dart.dev/tools/pub/publishing#what-files-are-published

mugikhan: We would still have to update the `.pubignore` to not ignore the binaries when publishing. Unless there is a better way to ignore them locally and not when publishing?

stevensJourney (dismissed):

rkistner (approved):

PR #92, opened 2024-06-04
Load and use PowerSync rust extension

mugikhan:
Continued from branch `rust-lib`

## Work done

- Load PowerSync rust extension using ffi
- Created `powersync_flutter_libs` package to download and bundle binaries

## TODO

- Publish `powersync_flutter_libs`

PR #151, opened 2024-08-22
chore(powersync): add max column check

DominicGBauer:
## Description
Users currently only see an error for maxmium number of columns at runtime. This adds a check so that they would see it on Table creation.

## Work Done
* Added checks to Table and added some Table specific tests


rkistner:
The schema is currently not validated in applications at all - appears to be a regression from #92 / 1.4.0. Could you please add that back? See https://github.com/powersync-ja/powersync.dart/pull/148/commits/4327b75d5f1cc2fe38d20b8f9a997ed27bf43c58 (also needs to be done for web).

This will also need a minor version bump at least.

DominicGBauer:
@rkistner this should now be ready to go

rkistner (approved):

Issue #166, opened 2024-09-18
Detect unsupported 32-bit environments


kobiebotha:
Running on an x86 Android emulator results in an error:

`'Unsupported processor architecture. X86 Android emulators are not supported. Please use an x86_64 emulator instead. All physical Android devices are supported including 32bit ARM.',`

The SDK is not catching this exception - it should, and then show a helpful error message to the user.

rkistner:
What is the difference between the current and expected behavior there? The message above already seems helpful to me

mugikhan:
> What is the difference between the current and expected behavior there? The message above already seems helpful to me

The message from the exception did not display to the user in the stacktrace. Addressed in #170 and will be fixed in the next release.

PR #170, opened 2024-09-19
Handle `PowersyncNotReadyException` printing the message for the user

mugikhan:
The SDK will now print the exception message for unsupported platforms when loading the core extension. The error message will state if the platform is unsupported. Addresses the issue reported in #166.

```
I/flutter ( 6406): [PowerSync] SEVERE: 2024-09-19 16:27:01.970790: Unsupported processor architecture. 
X86 Android emulators are not supported. 
Please use an x86_64 emulator instead. All physical Android devices are supported including 32bit ARM.
```

## Work done
- Print exception message when loading extension
- Remove unused imports

### DIFF
--- a/packages/powersync/lib/src/open_factory/native/native_open_factory.dart
+++ b/packages/powersync/lib/src/open_factory/native/native_open_factory.dart
@@ -71,6 +71,7 @@ class PowerSyncOpenFactory extends AbstractPowerSyncOpenFactory {
       enableExtension();
     } on PowersyncNotReadyException catch (e) {
       autoLogger.severe(e.message);
+      rethrow;
     }
 
     var db = super.open(options);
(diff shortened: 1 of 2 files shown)


packages/powersync/lib/src/open_factory/native/native_open_factory.dart
```diff
@@ -66,7 +67,11 @@ class PowerSyncOpenFactory extends AbstractPowerSyncOpenFactory {
     // ignore: deprecated_member_use_from_same_package
     _sqliteSetup?.setup();
 
-    enableExtension();
+    try {
+      enableExtension();
+    } on PowersyncNotReadyException catch (e) {
+      autoLogger.severe(e.message);
```
rkistner, line 73: This should probably rethrow the error, instead of continuing and failing with another error later on?

rkistner (approved):

PR #171, opened 2024-09-23
fix: propagate maxReaders parameter in PowerSyncDatabase.withFactory

davidmartos96:
The maxReaders parameter wasn't being passed down to the sqlite_async library, so always 5 readers were being used regardless of the value.


mugikhan:
Thanks for reporting this and creating a PR @davidmartos96, we have fixed and released this change in #181.

PR #172, opened 2024-09-25
[Feature] Add warning if crud transactions are not completed

mugikhan:
## Description

This is a backport of a feature added to the powersync js SDK. It ensures that there is a warning displayed when the same `CrudEntry` is processed by the upload loop. See this [PR](https://github.com/powersync-ja/powersync-js/pull/254) in the JS SDK for a more detailed explanation.

## Work done

- Display warning if crud transactions are in queue to be uploaded

### DIFF
--- a/packages/powersync/lib/src/streaming_sync.dart
+++ b/packages/powersync/lib/src/streaming_sync.dart
@@ -164,49 +164,52 @@ class StreamingSyncImplementation {
   }
 
   Future<void> uploadAllCrud() async {
-    // Keep track of the first item in the CRUD queue for the last `uploadCrud` iteration.
-    CrudEntry? checkedCrudItem;
+    return crudMutex.lock(() async {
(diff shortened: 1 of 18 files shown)


packages/powersync/lib/src/streaming_sync.dart
```diff
@@ -159,40 +164,49 @@ class StreamingSyncImplementation {
...
         await Future.delayed(retryDelay);
-      }
-    }
-    _updateStatus(uploading: false);
-  }
-
-  Future<bool> uploadCrudBatch() async {
-    return crudMutex.lock(() async {
-      if ((await adapter.hasCrud())) {
-        _updateStatus(uploading: true);
```
stevensJourney, line 181: We should not remove the `crudMutex` locking here. We should also keep the status update for `uploading: true`

stevensJourney (dismissed):
Looks good to me

stevensJourney (dismissed):

stevensJourney (approved):

PR #175, opened 2024-09-26
chore: update podlock

DominicGBauer:
## Description
Podfile.lock is outdated which results in an error on install


rkistner (approved):

rkistner:
Is it feasible to test the build on github actions to pick up issues like this earlier?

DominicGBauer:
@rkistner we should probably do something similar to this from JS repo https://github.com/powersync-ja/powersync-js/blob/main/.github/workflows/test-isolated.yaml

PR #193, opened 2024-10-16
Benchmark App

rkistner:
Add a benchmark app to test initial sync times and incremental sync latency.

Uses hardcoded temporary tokens, and [this backend](https://github.com/powersync-ja/powersync-nodejs-backend-todolist-demo).

image

See the [readme](https://github.com/powersync-ja/powersync.dart/blob/2017056c581ce82abdbc78294f6e60b67105f204/demos/benchmarks/README.md) for details.


demos/benchmarks/README.md
```diff
@@ -0,0 +1,86 @@
...
+INSERT INTO bulk_data (name, size_bucket) SELECT repeat('a', 20), '100k' FROM generate_series(1, 100000);
+INSERT INTO bulk_data (name, size_bucket) SELECT repeat('a', 20), '1m' FROM generate_series(1, 1000000);
+INSERT INTO bulk_data (name, size_bucket) SELECT repeat('a', 20), '10m' FROM generate_series(1, 10000000);
+```
+
+For reference, these rows are around 142 bytes each when synced. However, sync performance is more related to the number of rows than the total data size, unless you have much larger rows.
+
+## PowerSync Service
+
+Use this sync rules:
```
benitav, line 38:
```suggestion
Use these sync rules:
```

demos/benchmarks/README.md
```diff
@@ -0,0 +1,86 @@
+# PowerSync Benchmark app
```
benitav, line 1: I don't have a strong opinion, but I'm thinking this might fit better in a tools/ directory, similar to what we have for the diagnostics app

rkistner: There's quite a bit of tooling in this repo specific to the demos folder, so it would take some refactoring to do that

demos/benchmarks/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+The backend is not required measuring initial sync, but is required for measuring latency.
+
+Use the backend here: https://github.com/powersync-ja/powersync-nodejs-backend-todolist-demo
+
+It should write to the same Postgres database as configured above.
+
+## Configure the app
+
+Generate a temporary token, and configure the credentials in `lib/app_config.dart`.
```
benitav, line 61:
```suggestion
Generate a [temporary token](https://docs.powersync.com/installation/authentication-setup/development-tokens#development-tokens), and configure the credentials in `lib/app_config.dart`.
```

demos/benchmarks/README.md
```diff
@@ -0,0 +1,86 @@
...
+## Postgres
+
+```sql
+CREATE TABLE bulk_data (id uuid primary key default gen_random_uuid(), created_at timestamptz not null default now(), name text, size_bucket text);
+CREATE TABLE benchmark_items(id uuid primary key, description text, client_created_at timestamptz not null, client_received_at timestamptz, server_created_at timestamptz not null default now());
+
+INSERT INTO bulk_data (name, size_bucket) SELECT repeat('a', 20), '10k' FROM generate_series(1, 10000);
+INSERT INTO bulk_data (name, size_bucket) SELECT repeat('a', 20), '100k' FROM generate_series(1, 100000);
+INSERT INTO bulk_data (name, size_bucket) SELECT repeat('a', 20), '1m' FROM generate_series(1, 1000000);
+INSERT INTO bulk_data (name, size_bucket) SELECT repeat('a', 20), '10m' FROM generate_series(1, 10000000);
```
stevensJourney, line 31: Should we add a statement for ensuring the publication exists?

rkistner: In general my goal here isn't to provide the full setup instructions for a powersync instance, but rather just document the bits that are specific to the benchmarks.

stevensJourney (approved):

benitav (approved):

PR #194, opened 2024-10-17
Feat: PowerSync encryption with SQLCipher

mugikhan:
## Description

This enables the use of database encryption with SQLCipher. It also includes a refactor into two new packages, `powersync_core` (Dart only) and `powersync_sqlcipher` in order to achieve this.
`powersync_core` includes all of the original `powersync` setup and functionality and exports it for the other packages to use.

 - `powersync_core` is now used by `powersync` which is a thin package that depends on `sqlite3_flutter_libs` for Flutter projects
 - `powersync_core` is also used by `powersync_sqlcipher` which is a thin package that depends on `sqlcipher_flutter_libs`.
 - `powersync_attachments_helper` also depends on `powersync_core` for its basic functionality.
 
 The reason for this refactor is so that we can make `PowerSync` compatible with both Dart and Flutter apps. It also allows us to enable encryption for users who can now install and import `powersync_sqlcipher`. Finally, it maintains the original PowerSync API for users using `powersync`.
 
## TODO

- [x] Update READMEs


melos.yaml
```diff
@@ -33,7 +33,7 @@ scripts:
 
   prepare:demos:
     description: Download SQLite3 wasm for demos
-    run: dart run powersync:setup_web --no-worker
+    run: dart run powersync_core:setup_web --no-worker
```
rkistner, line 36: Should this rather stay in the `powersync` package?

1. Existing applications may depend on having `dart run powersync:setup_web` available.
2. The downloaded WASM files are specific to plain SQLite, and powersync_sqlcipher (if we ever support that on web) would need to use different files.

packages/powersync_core/README.md
```diff
@@ -0,0 +1,39 @@
+<!--
```
rkistner, line 1: Readme should be updated.

packages/powersync_core/NOTICE
```diff
@@ -0,0 +1 @@
+Copyright 2023 Journey Mobile, Inc.
```
rkistner, line 1:
```suggestion
Copyright 2024 Journey Mobile, Inc.
```

packages/powersync_sqlcipher/NOTICE
```diff
@@ -0,0 +1 @@
+Copyright 2023 Journey Mobile, Inc.
```
rkistner, line 1:
```suggestion
Copyright 2024 Journey Mobile, Inc.
```

packages/powersync_sqlcipher/lib/src/sqlcipher.dart
```diff
@@ -0,0 +1,39 @@
...
+      {required super.path, required this.key, super.sqliteOptions});
+
+  final String key;
+
+  @override
+  List<String> pragmaStatements(SqliteOpenOptions options) {
+    final basePragmaStatements = super.pragmaStatements(options);
+    return [
+      // Set the encryption key as the first statement
+      "PRAGMA KEY = '$key'",
```
rkistner, line 20:
```suggestion
      "PRAGMA KEY = ${quoteString(key)}",
```

This makes sure there is no SQL injection possibility if user input is directly used as the key. The `quoteString` function is part of sqlite_async.

melos.yaml
mugikhan: My original thought was that we could reuse it for SQLCipher but it isn't web compatible at this point so I moved it back into the `powersync` package.

packages/powersync_attachments_helper/example/getting_started.dart
```diff
@@ -1,7 +1,7 @@
 import 'dart:async';
 
 import 'package:powersync_attachments_helper/powersync_attachments_helper.dart';
-import 'package:powersync/powersync.dart';
+import 'package:powersync_core/powersync_core.dart';
```
stevensJourney, line 4: Is this the recommended approach for this package? Would it not be imported from the `powersync` or SQLCipher version package?

packages/powersync_sqlcipher/example/main.dart
```diff
@@ -0,0 +1,51 @@
...
+  if (!kIsWeb) {
+    final dir = await getApplicationSupportDirectory();
+    path = join(dir.path, 'powersync-dart.db');
+  }
+
+  // Setup the database.
+  final cipherFactory = PowerSyncSQLCipherOpenFactory(
+      path: path, key: "sqlcipher-encryption-key");
+
+  db = PowerSyncDatabase.withFactory(cipherFactory, schema: schema);
```
stevensJourney, line 40: Is it possible to extend the regular `PowerSyncDatabase` constructor with an option for the cipher key? That might make the API a bit easier to use.

packages/powersync_attachments_helper/example/getting_started.dart
rkistner: For a typical app you'd use one of those two. If you need to be compatible with both (like this helper lib), you'd use `powersync_core`.

mugikhan: I can add a comment showing that you can use `powersync` or `powersync_sqlcipher` in a client app?

stevensJourney: I see. I didn't scan through the entire file. I thought this example snippet actually instantiated a PowerSync client. I see it's just importing the type from the core package.

packages/powersync_sqlcipher/example/main.dart
mugikhan: As discussed, the current `PowerSyncDatabase` class only has factory constructors which doesn't allow it to be extended in Dart.

stevensJourney (dismissed):

packages/powersync/README.md
```diff
@@ -2,14 +2,20 @@
...
-This package (`powersync`) is the PowerSync client SDK for Dart/Flutter.
+This package (`powersync`) is the PowerSync client SDK for Flutter.
 
 See a summary of features [here](https://docs.powersync.com/client-sdk-references/flutter).
 
+> **Note**
+> This SDK is recommended for Flutter applications that do not require on-device encryption.
+
+> - For Flutter applications that require encryption, use the [`powersync_sqlcipher`](https://pub.dev/packages/powersync_sqlcipher) SDK.
+> - For use cases such as server-side Dart or non-Flutter Dart environments, use the [`powersync_core`](https://pub.dev/packages/powersync_core) SDK.
```
rkistner, line 17:
```suggestion
> This package is recommended for Flutter applications that do not require on-device encryption.

> - For Flutter applications that require encryption, use the [`powersync_sqlcipher`](https://pub.dev/packages/powersync_sqlcipher) package.
> - For use cases such as server-side Dart or non-Flutter Dart environments, use the [`powersync_core`](https://pub.dev/packages/powersync_core) package.
```

packages/powersync_core/pubspec.yaml
```diff
@@ -0,0 +1,43 @@
+name: powersync_core
+version: 0.0.1
```
rkistner, line 2: Can we start at a higher version, e.g. 1.0.0? While it's technically a new package, it was based on the powersync one, which is already stable.

packages/powersync_sqlcipher/pubspec.yaml
```diff
@@ -0,0 +1,31 @@
+name: powersync_sqlcipher
+version: 0.0.1
```
rkistner, line 2: Can we start at `0.1.0`? I generally view 0.0.x versions as "very experimental".

rkistner (commented):
The melos.yml `version` -> `preCommit` command needs to be updated - the version.dart file has moved.

packages/powersync_core/pubspec.yaml
mugikhan: Yes I will run melos version and bump the version to a stable 1.0.0, once any other issues are resolved.

packages/powersync_sqlcipher/pubspec.yaml
mugikhan: Will do with melos version once the issues are resolved.

mugikhan:
> The melos.yml `version` -> `preCommit` command needs to be updated - the version.dart file has moved.

Should I move the version file back to the `powersync` package, it would live in the lib folder since it refers to that libraries version and not `powersync_core`?

rkistner (dismissed):

rkistner (approved):

PR #203, opened 2024-10-30
[Fix] CRUD Upload on Reconnect

stevensJourney:
# Overview

This fixes an issue where CRUD uploads are not triggered when the SDK reconnects to the PowerSync service after being offline.

Previously uploads were only triggered when an initial call to `.connect` has been made or when changes to the `ps_crud` table have occurred. If an application called `.connect` successfully, then went offline and made CRUD operations - the SDK would not trigger a CRUD upload once it had re-established a connection.

This now triggers a CRUD upload whenever the SDK establishes a connection or a change has been made to `ps_crud`.

## Testing

This was tested locally and unit tests have been added using a mock PowerSync service.

### DIFF
--- a/packages/powersync/test/streaming_sync_test.dart
+++ b/packages/powersync/test/streaming_sync_test.dart
@@ -13,8 +13,11 @@ final testUtils = TestUtils();
 
 class TestConnector extends PowerSyncBackendConnector {
   final Function _fetchCredentials;
+  final Future<void> Function(PowerSyncDatabase)? _uploadData;
 
-  TestConnector(this._fetchCredentials);
+  TestConnector(this._fetchCredentials,
+      {Future<void> Function(PowerSyncDatabase)? uploadData})
+      : _uploadData = uploadData;
 
   @override
   Future<PowerSyncCredentials?> fetchCredentials() {
@@ -22,7 +25,11 @@ class TestConnector extends PowerSyncBackendConnector {
   }
 
   @override
-  Future<void> uploadData(PowerSyncDatabase database) async {}
+  Future<void> uploadData(PowerSyncDatabase database) async {
+    if (_uploadData != null) {
+      await _uploadData(database);
+    }
+  }
 }
 
 void main() {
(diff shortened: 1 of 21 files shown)


packages/powersync/test/streaming_sync_test.dart
```diff
@@ -13,16 +13,23 @@ final testUtils = TestUtils();
...
   Future<PowerSyncCredentials?> fetchCredentials() {
     return _fetchCredentials();
   }
 
   @override
-  Future<void> uploadData(PowerSyncDatabase database) async {}
+  Future<void> uploadData(PowerSyncDatabase database) async {
+    if (_uploadData != null) {
+      await _uploadData(database);
+    }
```
mugikhan, line 31:
```suggestion
    await _uploadData?.call(database);
```

mugikhan (dismissed):
LGTM!

rkistner:
For my own reference, since I was confused with how this could happen:
1. If a crud uploads fails with just an upload error, it is automatically retried indefinitely.
2. If the user explicitly disconnects and reconnects, the crud loop is restarted, and should try again.
3. If however a crud upload fails and `isConnected == false`, then the current crud iteration stops, and waits for more changes. This prevents continuous retrying of uploads while the user is offline.

The fix here targets (3), by basically triggering the "there are more changes" stream when the sync connection is established again.

lucastex:
@stevensJourney thanks for being working on this. We're following the PR because it got a huge impact in our last update here.  Thanks.

stevensJourney:
> Some unit tests are broken. Will investigate.

Tests pass more consistently now. The main issues were from the `full powersync reconnect` test timing out: this test calls the SDK `close` method which performs a disconnect and closes the SQLite DB. A race condition caused the SQLite DB to close before the CRUD loop exited. This caused an exception to be thrown when the CRUD loop checked for CRUD operations - the caught exception added a retry delay which kept the sync isolate alive for an additional 5 seconds. 

I've shuffled around the CRUD loop exit checks in order to exit the loop faster when disconnecting or closing.

rkistner (dismissed):

mugikhan (dismissed):

mugikhan (approved):

PR #206, opened 2024-11-04
Fix multi-tab sync issues

rkistner:
1. Fix client parameters not being passed to the server (regression from #200).
 2. Reconnect when a client calls `connect()` again with different parameters.
 3. Immediately disconnect when any client calls `disconnect()` (or `disconnectAndClear()`).

(2) is to make sure we connect with the latest client parameters. I'm not sure if it's perhaps better to _always_ reconnect when calling `connect()` - that's what we do in other SDKs (including native Flutter). However, that may not be optimal when working with multiple tabs, since it would effectively reconnect every time you open a new tab.

The last one is specifically an issue when calling `disconnectAndClear()`. If other tabs were open, the connection would remain open, despite deleting all data in the database. This caused the app to end up in a weird state.

### DIFF
--- a/packages/powersync/lib/src/database/web/web_powersync_database.dart
+++ b/packages/powersync/lib/src/database/web/web_powersync_database.dart
@@ -148,11 +148,11 @@ class PowerSyncDatabaseImpl
     // duplicating work across tabs.
     try {
       sync = await SyncWorkerHandle.start(
-        this,
-        connector,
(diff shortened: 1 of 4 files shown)


stevensJourney:
FWIW We currently reconnect on every call to `connect` in the JS Web SDK, that does cause a reconnect whenever a new tab is opened.

stevensJourney (approved):

PR #207, opened 2024-11-05
Simplify Flutter Web Readme

benitav:
Most content will be moved to [docs](https://docs.powersync.com/client-sdk-references/flutter/flutter-web-support), so only keep the essential getting started instructions in the Readme.

Depends on https://github.com/powersync-ja/mintlify-docs/pull/17/files


stevensJourney (approved):
Happy from my side with this.

PR #221, opened 2024-12-24
feat: add trello demo

DominicGBauer:
## Description
This includes the trello demo into dart monorepo. The demo has been updated to only have a single demo app as opposed to an app and a powersync client. The reason for this is to make it conform to the other demos and is most likely how a user would use the powersync SDK. Some changes needed to take place as a lot of the dependencies needed to be upgraded.

There do appear to be some UI issues which will need to be addressed later (I think these have been there for some time).

## Testing
https://github.com/user-attachments/assets/6b1b9267-7c21-43e9-81e3-148b89223c79


Chriztiaan (commented):
I am happy with the PR. Seeing as it's mostly copied over code I just want to test it myself by going through the readme - will do that tomorrow.

Chriztiaan:
Tested, looks good.

Chriztiaan (approved):

PR #232, opened 2025-01-27
Update lints and dependencies in released packages

simolus3:
In `package:powersync_core`, migrate away from `package:dcli` and `package:js` so that we're compatible with the latest version of all dependencies.
Apart from that, this updates `package:lints` and fixes the new analysis warnings.


stevensJourney (approved):
Looks good to me :)

Issue #137, opened 2024-08-02
Validate Fields Early


mcmah309:
I got into a state where
https://github.com/powersync-ja/powersync.dart/blob/512357877a50d1ac0b7c32ea08ce302336c94a51/packages/powersync/lib/src/crud.dart#L86
was failing since somehow `data['id']` was an `int`. I didn't see an easy way to resolve this so I had to run 
```dart
await db.execute("DELETE FROM ps_crud;");
```
To drop the queue. I believe this resulted from me accidentally writing something like
```sql
INSERT INTO table_name(id, ...) VALUES (random_int)
```
Instead of
```sql
INSERT INTO table_name(id, ...) VALUES (uuid())
```
Which messed up all of syncing, even if I deleted the original code/table. Ideally a mistake like this would never make it to production, but if it did, is there a better way to recover without losing user data? Or at least in this case, validate early so for essential fields these situations cannot occur.

It is worth noting that all subsequent inserts would not sync until `ps_crud` was dropped.

rkistner:
Thanks for reporting the issue. I've seen at least one other report of the same issue, and we'll implement a validation for this.

> is there a better way to recover without losing user data?

You could probably replace the ids in the `ps_crud` table if you need to keep the data, e.g. `update ps_crud set id = cast(id as text)`. I do imagine this issue should be picked up before production in most cases though, since it will affect all writes to that table.

PR #238, opened 2025-02-02
Validate data['id'] field early in dev

LucDeCaf:
Implements change proposed in https://github.com/powersync-ja/powersync.dart/issues/137.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,40 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2025-01-28
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+ - There are no breaking changes in this release.
+
+Packages with other changes:
+
+ - [`powersync_core` - `v1.1.2`](#powersync_core---v112)
+ - [`powersync_attachments_helper` - `v0.6.18`](#powersync_attachments_helper---v0618)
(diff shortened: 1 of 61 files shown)


simolus3 (commented):
Thanks for the contribution!

In my opinion, we should add explicit casts to all our JSON-deserializing constructors (and also enable [strict casts](https://dart.dev/tools/analysis#enabling-additional-type-checks) to check this).
`dart2js` with `-O3` or `-O4` [can omit implicit casts](https://dart.dev/tools/dart-compile#basic-options) such as the ones in our `fromRow` methods, which leads to unsound code when the type in the JSON value is not what we expect (it leads to `num` values creeping into `String` fields, possibly breaking type safety everywhere).
The general advice is to not compile with `-O3` without carefully checking for implicit casts, but:

1. We compile our web workers with `-O4`.
2. All Flutter web apps are compiled with `-O4` (a _very_ dangerous decision by Flutter, but we have to deal with that).

So ideally, `CrudEntry.fromRow` should look like this:

```dart
final data = jsonDecode(row['data'] as String);
return CrudEntry(
  row['id'] as int,
  UpdateType.fromJsonChecked(data['op'] as String),
  data['type'] as String,
  data['id'] as String,
  row['tx_id'] a String,
  data['data'],
);
```

There are a few methods affected by this which should all be fixed. That will be a bigger change, so you can leave that to us if you don't want to patch them all :D

LucDeCaf:
@simolus3 Thank you for the insight! I'm renaming the branch to reflect the new changes so expect a new PR soon ;)

PR #239, opened 2025-02-03
Validate input for CrudEntry.fromRow explicitly

LucDeCaf:
As suggested in #137.

### DIFF
--- a/packages/powersync_core/analysis_options.yaml
+++ b/packages/powersync_core/analysis_options.yaml
@@ -13,6 +13,12 @@
 
 include: package:lints/recommended.yaml
 
+analyzer:
+  language:
+    strict-casts: true
+    strict-inference: true
+    strict-raw-types: true
+
 # Uncomment the following section to specify additional rules.
 
 # linter:
--- a/demos/benchmarks/pubspec.lock
+++ b/demos/benchmarks/pubspec.lock
@@ -305,21 +305,21 @@ packages:
       path: "../../packages/powersync"
       relative: true
     source: path
-    version: "1.11.1"
+    version: "1.10.0"
   powersync_core:
     dependency: "direct overridden"
     description:
       path: "../../packages/powersync_core"
       relative: true
     source: path
-    version: "1.1.1"
+    version: "1.0.0"
   powersync_flutter_libs:
     dependency: "direct overridden"
     description:
       path: "../../packages/powersync_flutter_libs"
       relative: true
     source: path
-    version: "0.4.4"
+    version: "0.4.3"
   pub_semver:
     dependency: transitive
     description:
--- a/demos/benchmarks/pubspec.lock
+++ b/demos/benchmarks/pubspec.lock
@@ -305,21 +305,21 @@ packages:
       path: "../../packages/powersync"
       relative: true
     source: path
-    version: "1.10.0"
+    version: "1.11.1"
   powersync_core:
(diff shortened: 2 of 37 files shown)


LucDeCaf:
@simolus3 Here's the new branch - sorry for the move, I'm a bit of a stickler for accurate branch names

rkistner:
I think this is not going to change the issue fundamentally - maybe the error will be a little clear, but in the reported scenario the user will still end up with an upload queue being "stuck".

What I meant by "validation" for this is we need to validate the type of the field when _writing_, which is when the mismatch occurs. The best place for that is probably in the triggers, such as [here](https://github.com/powersync-ja/powersync-sqlite-core/blob/024fda8ede50a4b0273d8f23e88ba9db996d371e/crates/core/src/views.rs#L155).

LucDeCaf:
Okay, will try that out and update PR later

rkistner:
I discussed with @simolus3 - The additional casts here are still good to have, we just _additionally_ need validations when writing.

LucDeCaf:
@rkistner Should I make a PR in the sqlite repo as well then?

LucDeCaf:
I have two ideas - either to `CAST` `NEW.id` to a string or to check if `NEW.id` matches a regex that corresponds to UUIDv4 (I believe that is the only supported format for Id's right now)

The regex I believe will work is this (assuming `powersync-sqlite-core` supports REGEXP):
```js
/[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}/
```

rkistner:
We support any TEXT ids, not only UUIDs (although we do recommend UUIDs for most use cases). I think validating that the id is TEXT is the best route here (probably using the sqlite typeof function), since automatically casting could cause unexpected behavior for developers down the line.

simolus3:
> I'm a bit of a stickler for accurate branch names

Then it looks like I've just ruined it by pushing applying the same principle to the entire `powersync_core` package.

I think I got the explicit casts right - unit tests pass and I've also tested these changes against the `supabase-todolist` demo with an updated worker.

rkistner:
> Right, I remember that somewhere in the back of my mind. I agree, auto-casting would be confusing - rather just raise an error.
> 
> I imagine something like this?
> 
> ```sql
> SELECT CASE
>       WHEN (NEW.id IS NULL)
>       THEN RAISE (FAIL, 'id is required')
>       WHEN (typeof(NEW.id) != 'text')
>       THEN RAISE (FAIL, 'id should be text')
>       END;
> ```

Yes, that should work

packages/powersync_core/analysis_options.yaml
```diff
@@ -13,6 +13,12 @@
 
 include: package:lints/recommended.yaml
 
+analyzer:
+  language:
+    strict-casts: true
+    strict-inference: true
+    strict-raw-types: true
```
rkistner, line 20: These looks like good checks to have - should we enable these on all our packages? Not the demos, but anything that we publish. Can be a separate PR.

rkistner (approved):

packages/powersync_core/analysis_options.yaml
simolus3: Agreed, I'll open a PR to follow up on this 👍

PR #243, opened 2025-02-10
Support bucket priorities and partial checkpoints

simolus3:
Add support for partial checkpoint complete messages, which are forwarded to the core extension by providing the names of completed buckets within that partial checkpoint.
This also adds extends the `waitForFirstSync` and `hasSynced` APIs to be aware of priorities.

These changes are not backwards-compatible with old versions of the extension (which would always report an inconsistent partial checkpoint), but the extension changes are backwards-compatible with other SDKs that don't support partial checkpoints.


rkistner:
> These changes are not backwards-compatible with old versions of the extension (which would always report an inconsistent partial checkpoint), but the extension changes are backwards-compatible with other SDKs that don't support partial checkpoints.

Could you update the version check here when the new version is released?
https://github.com/powersync-ja/powersync.dart/blob/1f9268fc42273ea58905f8f8e2ef57b52a077ff3/packages/powersync_core/lib/src/database/powersync_db_mixin.dart#L84

packages/powersync_core/lib/src/database/powersync_db_mixin.dart
```diff
@@ -121,27 +121,65 @@ mixin PowerSyncDatabaseMixin implements SqliteConnection {
...
+        statusInPriority: priorityStatus,
+      );
       setStatus(status);
     }
   }
 
-  /// Returns a [Future] which will resolve once the first full sync has completed.
-  Future<void> waitForFirstSync() async {
-    if (currentStatus.hasSynced ?? false) {
+  /// Returns a [Future] which will resolve once a synchronization operation has
```
benitav, line 162: "sync operation" is confusing since in our usage and pricing terminology it refers to a single row/operation that transfers from the service to a client. In this case I think we're referring to  a sync "cycle" as a whole? I think we generally just use the term "sync" for this, but Steven/Ralf might know better.

demos/supabase-todolist/README.md
```diff
@@ -29,6 +29,25 @@ Create a new PowerSync instance, connecting to the database of the Supabase proj
...
+    data:
+      - select * from lists where id = bucket.list_id
+
+  user_todos:
+    parameters: select id as list_id from lists where owner_id = request.user_id()
+    data:
+      - select * from todos where list_id = bucket.list_id
+```
+
+The rules synchronize list with a higher priority the items within the list. This can be
```
benitav, line 46:
```suggestion
These rules showcase [prioritized sync](https://docs.powersync.com/usage/use-case-examples/prioritized-sync), by syncing a user's lists with a higher priority than the items within a list (todos). This can be
```

stevensJourney (approved):

PR #247, opened 2025-03-03
Release

simolus3:
First, I've fixed a weird issue we had when running demos with the DDC compiler through `flutter run`. After a hot reloads, we got weird type errors because the list of partial priority entries suddenly had an unexpected type. Given that it only happens after reloads and only on the web, I'll call this a compiler bug. I was unable to reproduce it in a smaller project, but I found a workaround that is applied here (making the `ListEquality` instance for `SyncStatus` more broad than necessary, and avoiding a constant list of entries when reading priority entries from workers).
I've also fixed a regression I've introduced in the supabase todolist example that showed the "Busy with sync" message precisely when it wasn't supposed to show...

Apart from that, this adds a minor version bump to `powersync_core`, `powersync` and `powersync_sqlcipher` to release those packages with support for bucket priorities.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,49 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2025-03-03
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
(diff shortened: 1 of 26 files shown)


stevensJourney (approved):
Looks good to me

stevensJourney (approved):

Issue #248, opened 2025-03-04
Version 1.12.0 Initialization error in web build


dh0er:
Hi,

i get the following error when trying to initialize ps database in my web build:

```
DartError: SqliteException(1): SqliteException(1): while preparing statement, no such table: ps_sync_state, SQL logic error (code 1)
  Causing statement: SELECT priority, last_synced_at FROM ps_sync_state ORDER BY priority;
```

I tried to delete the db so it gets recreated, but no effect. Version 1.11.3 is still working.

Thanks

simolus3:
Thanks for the report! Could you run `dart run powersync:setup_web` again (after upgrading to 1.12.0) to update the workers and WebAssembly assets?

dh0er:
That worked :)
Thanks for the quick support!

PR #249, opened 2025-03-05
Improve compatibility check for core extension

simolus3:
When opening databases, we check whether the version of the core extension we've loaded is compatible with our Dart SDK. This check was not updated for the update to bucket priorities, allowing SDK versions `<0.3.11` to be used. This breaks immediately afterwards when selecting from `ps_sync_state`.

This updates the version check to the new minimum version and refactors the check to make minimum and maximum versions easier to update the future.
When the version check fails on the web (the most likely cause since we expect `powersync` and `powersync_flutter_libs` to be updated as one, but require users to update our WebAssembly files manually), this also improves the error message to point to the docs explaining how to update these assets.

Resolves https://github.com/powersync-ja/powersync.dart/issues/248.

### DIFF
--- a/packages/powersync_core/lib/powersync_core.dart
+++ b/packages/powersync_core/lib/powersync_core.dart
@@ -9,7 +9,6 @@ export 'src/database/powersync_database.dart';
 export 'src/exceptions.dart';
 export 'src/log.dart';
 export 'src/open_factory.dart';
-export 'src/powersync_database.dart';
 export 'src/schema.dart';
 export 'src/sync_status.dart';
 export 'src/uuid.dart';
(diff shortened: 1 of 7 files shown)


packages/powersync_core/lib/powersync_core.dart
```diff
@@ -9,7 +9,6 @@ export 'src/database/powersync_database.dart';
 export 'src/exceptions.dart';
 export 'src/log.dart';
 export 'src/open_factory.dart';
-export 'src/powersync_database.dart';
```
simolus3, line 12: This file just exports `database/powersync_database.dart`, which is already exported here.

rkistner (approved):

rkistner:
Just need to update the tests to use the new extension

rkistner (approved):

PR #251, opened 2025-03-10
Fix broken token invalidation on web

simolus3:
This fixes the issue [reported here](https://discord.com/channels/1138230179878154300/1344044450191773746). When the sync service calls `invalidCredentialsCallback`, the app should fetch new credentials which are then cached by the connector. For this, we need to call `prefetchCredentials`. `fetchCredentials`, the method we used to call, obtains credentials without saving them.

We were already calling `prefetchCredentials` on native platforms, but the web implementation was broken and never refreshed the token. This caused us to keep using the expired token, entering an infinite loop of 401 errors.

### DIFF
--- a/packages/powersync_core/test/in_memory_sync_test.dart
+++ b/packages/powersync_core/test/in_memory_sync_test.dart
@@ -20,9 +20,11 @@ void main() {
     late CommonDatabase raw;
     late PowerSyncDatabase database;
     late MockSyncService syncService;
-    late StreamingSyncImplementation syncClient;
+    late StreamingSync syncClient;
+    var credentialsCallbackCount = 0;
 
     setUp(() async {
+      credentialsCallbackCount = 0;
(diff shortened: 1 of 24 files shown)


packages/powersync_core/test/in_memory_sync_test.dart
```diff
@@ -312,18 +316,37 @@ void main() {
         },
       );
     });
+
+    test('reconnects when token expires', () async {
```
simolus3, line 320: This doesn't actually test the fix because it only affects the channel between the sync worker and the database.

I want to revisit this in the future and run these tests with a fake worker based on a `MessageChannel` in JS. That will take some more work though, so I've added the test now and will do the refactoring in the future.

stevensJourney (dismissed):
Looks good to me

stevensJourney (approved):

PR #257, opened 2025-03-21
Demos: Replace `setState` with `StreamBuilder` where appropriate

simolus3:
Some pages in the demo apps create a stream subscription and use `setState()` when receiving new snapshots to render content. This refactors most of these usages to use Flutter's `StreamBuilder` widget. Using stream builders makes it impossible to leak resources (the widget will stop listening to the stream when removed from the tree) and requires much less boilerplate code.


rkistner (approved):
Thanks, this is much cleaner!

PR #259, opened 2025-03-24
Drift demo with riverpod

simolus3:
This changes the `supabase_todolist_drift` package to also use Riverpod for state management. This example can then serve as a guide for integrating PowerSync with popular state management solutions (I'll open a PR to our documentation linking towards it).

The idea is to adopt best practices of modern state management in Flutter. This changes the example to:

- Extract the PowerSync database and relevant fields (like the auto-updating `SyncState`) into Riverpod providers.
- Consume providers from widgets (for simple queries) or through wrapping `Notifier`s to bundle similar functionality.
- Adopt [auto_route](https://pub.dev/packages/auto_route) for declarative routing.


stevensJourney (approved):
I don't have much context on the best practices of state management in Flutter/Dart, but from a general architecture perspective this looks brilliant. Happy with these changes.

PR #260, opened 2025-03-24
Sync progress status

simolus3:
This extends `SyncStatus` to report download progress as it happens. Progress is reported as a `(completed, total, fraction)` triple (where `fraction = completed / total`) and available:

1. For entire sync operations: `db.currentStatus.downloadProgress.untilCompletion`.
2. For the progress until a specific priority: `db.currentStatus.downloadProgress.untilPriority(myPrio)`.

The best way to get a look and feel for the API may be to look at this example usage:

https://github.com/powersync-ja/powersync.dart/blob/9d45ba37bfaff529246c7124231bcfcbab2613fb/demos/supabase-todolist/lib/widgets/guard_by_sync.dart#L26-L53

In an actual app, it looks like this:

https://github.com/user-attachments/assets/fa3ed419-a47f-4321-b59d-b2d1d512d513

There is some delay in the `sync_local` commit not included in the progress bar now, but the amount of operations is updated as it should be.


packages/powersync_core/lib/src/sync_status.dart
```diff
@@ -202,3 +215,170 @@ class UploadQueueStats {
...
+  final int _totalTarget;
+
+  InternalSyncDownloadProgress(this.downloaded, this.target)
+      : _totalDownloaded = target.values.sum,
+        _totalTarget = target.values.sum;
+
+  factory InternalSyncDownloadProgress.fromZero(Checkpoint target) {
+    final targetOps = target.checksums.groupFoldBy<BucketPriority, int>(
+      (cs) => BucketPriority(cs.priority),
+      (prev, cs) => (prev ?? 0) + (cs.count ?? 0),
```
rkistner, line 234: Note that the count here is the _total_ count for the bucket, not the incremental count between checkpoints. We'll need to add a new field to the API to include the count since the last checkpoint.

packages/powersync_core/lib/src/sync_status.dart
```diff
@@ -202,3 +215,170 @@ class UploadQueueStats {
     }
   }
 }
+
+@internal
+final class InternalSyncDownloadProgress {
+  final Map<BucketPriority, int> downloaded;
+  final Map<BucketPriority, int> target;
```
rkistner, line 222: It may be more accurate if we keep track of the counts for each specific bucket, rather than per priority.

In _most_ cases it shouldn't make a difference, but there are some edge cases where it does:
1. A checkpoint may be "interrupted" before all data is sent. The new checkpoint after the interruption may add buckets, remove buckets, or change the priority of buckets.
2. The counts are not guaranteed to be 100% accurate in all cases. There are some cases such as compacting buckets, which may alter the count without affecting the integrity - potentially while the client is downloading data. To cater for this, we may want to enforce that:
    a. `downloaded <= target` for each bucket (never go over 100%).
    b. `downloaded = target` when we have all data for the bucket.

rkistner (commented):
I'm happy with the overall structure & apis - just added some implementation comments.

demos/supabase-todolist/lib/widgets/guard_by_sync.dart
```diff
@@ -0,0 +1,55 @@
+import 'package:flutter/material.dart';
+import 'package:powersync/powersync.dart' hide Column;
+import 'package:powersync_flutter_demo/powersync.dart';
+
+/// A widget that shows [child] after a complete sync on the database has
+/// completed and a progress bar before that.
+class GuardBySync extends StatelessWidget {
```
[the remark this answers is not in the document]
simolus3: I think this is a good idea 👍 Given that we already have `package:powersync` as a Flutter-only library, I think it could reasonably live there (or at least be exported from there, we might want a `package:powersync_widgets` to share code between the main package and the SQLCipher one). It just can't be in `package:powersync_core` so that we can keep using the package without Flutter.

packages/powersync_core/lib/src/sync_status.dart
simolus3: We're now tracking operations on the client-side instead.

packages/powersync_core/lib/src/sync/sync_status.dart
```diff
@@ -0,0 +1,377 @@
...
+  /// download is complete.
+  final int totalOperations;
+
+  /// How many operations have already been downloaded since the last complete
+  /// download.
+  final int downloadedOperations;
+
+  ProgressWithOperations._(this.totalOperations, this.downloadedOperations);
+
+  /// Relative progress (as a number between `0.0` and `1.0`).
```
rkistner, line 337: Maybe explicitly note there could be a progress of 1.0 even though we're busy saving the data.

packages/powersync_core/lib/src/sync/sync_status.dart
```diff
@@ -0,0 +1,377 @@
...
+    return 'for total: ${all.downloadedOperations} / ${all.totalOperations}';
+  }
+
+  static const _mapEquality = MapEquality<Object?, Object?>();
+}
+
+/// Information about a progressing download.
+///
+/// This reports the `total` amount of operations to download, how many of them
+/// have alreaady been `completed` and finally a `fraction` indicating relative
```
rkistner, line 321:
```suggestion
/// have already been `completed` and finally a `fraction` indicating relative
```

rkistner (commented):
I didn't check the details of the sync status refactoring, but I'm happy with the overall approach.

demos/supabase-todolist/lib/widgets/guard_by_sync.dart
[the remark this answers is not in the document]
simolus3: I've moved this into a separate task so that we can ship this first - I think coming up with a good design and API for the progress widgets deserves another PR.

rkistner (approved):

PR #263, opened 2025-04-04
Fix race condition for checkpoints during uploads

simolus3:
This is the Dart port of https://github.com/powersync-ja/powersync-js/pull/558, see that PR for details.

### DIFF
--- a/packages/powersync_core/lib/src/streaming_sync.dart
+++ b/packages/powersync_core/lib/src/streaming_sync.dart
@@ -136,7 +136,7 @@ class StreamingSyncImplementation implements StreamingSync {
     try {
       _abort = AbortController();
       clientId = await adapter.getClientId();
-      crudLoop();
+      _crudLoop();
       var invalidCredentials = false;
       while (!aborted) {
         _updateStatus(connecting: true);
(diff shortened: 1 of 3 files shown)


packages/powersync_core/lib/src/streaming_sync.dart
```diff
@@ -244,7 +247,11 @@ class StreamingSyncImplementation implements StreamingSync {
           _updateStatus(uploading: false);
         }
       }
-    }, timeout: retryDelay);
+    }, timeout: retryDelay).whenComplete(() {
+      assert(identical(_activeCrudUpload, completer));
```
stevensJourney, line 251: For my understanding: Under what circumstances would this assertion fail and how would we recover from this?

simolus3: Hopefully, in no circumstances at all. We assert that there aren't two concurrent calls to `_uploadAllCrud()` (which is aready ensured by our code because `_uploadAllCrud()` is only called by `crudLoop`, which in turn is only called once in `streamingSync()`). I just wanted to spell that assumption out in code.

stevensJourney (approved):

PR #267, opened 2025-04-15
Refactor connection locks

simolus3:
The existing implementation around `connect()` and `disconnect()` has a bit of an odd control flow. I couldn't find any likely race conditions, but this still cleans up the implementation a bit to be easier to reason about. In particular, some of the things this addresses:

1. While we acquire a mutex for `connect()`, we were not doing this for `disconnect()`.  So if we call `connect()` and then immediately call `disconnect()`, it's possible that `disconnect()` does nothing because the `AbortController` has not yet been set.
2. `reconnect()` doesn't actually check if we still need to connect, so a `disconnect()` call followed by an error on the isolate could cause another `connect()` that is not stopped.
3. We set an `onExit` port when spawning the isolate, but there's a separate `Isolate.current.addOnExitListener` command in the background isolate. The two trigger different behavior, this has now been cleaned up.
4.  The `Isolate.spawn` call is not awaited, meaning that another `connect()` call could begin before the earlier one started setting up the sync client. I don't think this causes any issues, but it's still better to restrict concurrency around `connect()`.

This also adds a lot of internal assertions to ensure we're not calling the internal connect methods concurrently. I'm still not sure how the problem with duplicate sync streams was caused though, the existing implementation should have guarded against that too.

### DIFF
--- a/packages/powersync_core/lib/src/database/native/native_powersync_database.dart
+++ b/packages/powersync_core/lib/src/database/native/native_powersync_database.dart
@@ -282,6 +282,8 @@ Future<void> _syncIsolate(_PowerSyncDatabaseIsolateArgs args) async {
   StreamSubscription<void>? localUpdatesSubscription;
 
   Future<void> shutdown() async {
+    await openedStreamingSync?.abort();
+
     localUpdatesSubscription?.cancel();
     db?.dispose();
     crudUpdateController.close();
@@ -291,9 +293,10 @@ Future<void> _syncIsolate(_PowerSyncDatabaseIsolateArgs args) async {
     // It needs to be closed before killing the isolate
     // in order to free the mutex for other operations.
     await mutex.close();
-    await openedStreamingSync?.abort();
-
     rPort.close();
+
+    // TODO: If we closed our resources properly, this wouldn't be necessary...
+    Isolate.current.kill();
   }
 
   rPort.listen((message) async {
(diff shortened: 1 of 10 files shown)


stevensJourney (dismissed):
Happy with the changes from my side.

packages/powersync_core/lib/src/database/native/native_powersync_database.dart
```diff
@@ -15,11 +15,12 @@ import 'package:powersync_core/src/log_internal.dart';
 import 'package:powersync_core/src/open_factory/abstract_powersync_open_factory.dart';
 import 'package:powersync_core/src/open_factory/native/native_open_factory.dart';
 import 'package:powersync_core/src/schema.dart';
-import 'package:powersync_core/src/schema_logic.dart';
 import 'package:powersync_core/src/streaming_sync.dart';
 import 'package:powersync_core/src/sync_status.dart';
 import 'package:sqlite_async/sqlite3_common.dart';
 import 'package:sqlite_async/sqlite_async.dart';
+// ignore: implementation_imports
+import 'package:sqlite_async/src/native/native_isolate_mutex.dart';
```
stevensJourney, line 23: Is this fine to ignore?

stevensJourney (approved):
The implementation looks solid from my side. I could not spot any glaring issues. I see there are some TODOs for the future, but they are documented nicely.

PR #269, opened 2025-04-18
Relax fileExtension requirement in attachmentQueue

### LINKED ISSUE #268: Attachments Helper – Allow Multiple File Extensions
Hi wonderful Powersync team 🙌

Following up on the discussion started here: https://discord.com/channels/1138230179878154300/1362146693457379572 (thanks again @benitav!), I’ve been experimenting with the Attachments helper and encountered a few bumps along the way.

**Use Case: Multiple file extensions in the same bucket**
I’d like to store files with various extensions in the same bucket (in Supabase, in my case). Currently, the library requires specifying a single fileExtension for each attachmentQueue, which limits us to one file type per queue, as it's passed to the watchIds method.

In my setup, I store the file extension directly in the filename on the bucket, so I don't need to append it again via the helper.

I’m not sure yet if this is a real requirement or if I’m approaching it the wrong way—but I’d love your thoughts 🤓

WDYT?

Thanks a lot, and have a great (syncing) day! ☀️

n0mad-d3v:
Hello powersync team 👋

This PR fixes #268 and updates the watchIds method in the Attachments helper to make the fileExtension parameter optional. This enables support for storing files with different extensions in the same attachmentQueue, which is useful in cases where:

- The extension is already part of the filename (e.g., managed directly in the storage bucket like Supabase).
- The developer wants more flexibility in handling mixed file types under a unified queue.

Motivation: Currently, the API assumes a single extension per queue, which limits use cases that involve multiple file types. Making fileExtension optional opens up broader usage while maintaining backward compatibility.

⚠️ This includes a breaking change. And along that, I'm wondering if it shouldn't be pushed further to be able to specify a Map<String, String?> for <filename, extension?> in the whatchIds to allow even more flexibility and usecases. 

This change was inspired by this [Discord thread](https://discord.com/channels/1138230179878154300/1362146693457379572).

WDYT? Looking forward to feedback!

PS: There is also a change about ensuring directory creation. In my usecase, I'm using Teams' ids as subdirectory so it's fully dynamic and my filename is like the following: 
`{teamId}/{attachmentType}/{attachmentId}.{attachmentExtension}`.

### DIFF
--- a/demos/supabase-todolist-drift/lib/attachments/queue.dart
+++ b/demos/supabase-todolist-drift/lib/attachments/queue.dart
@@ -68,7 +68,7 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
   }
 
   @override
-  StreamSubscription<void> watchIds({String fileExtension = 'jpg'}) {
+  StreamSubscription<void> watchIds({String? fileExtension}) {
     log.info('Watching photos in $todosTable...');
     return db.watch('''
       SELECT photo_id FROM $todosTable
--- a/demos/supabase-todolist/lib/attachments/queue.dart
+++ b/demos/supabase-todolist/lib/attachments/queue.dart
@@ -68,7 +68,7 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
   }
 
   @override
-  StreamSubscription<void> watchIds({String fileExtension = 'jpg'}) {
+  StreamSubscription<void> watchIds({String? fileExtension}) {
     log.info('Watching photos in $todosTable...');
     return db.watch('''
       SELECT photo_id FROM $todosTable
--- a/packages/powersync_attachments_helper/README.md
+++ b/packages/powersync_attachments_helper/README.md
@@ -62,7 +62,7 @@ class PhotoAttachmentQueue extends AbstractAttachmentQueue {
   // This watcher will handle adding items to the queue based on
   // a users table element receiving a photoId
   @override
-  StreamSubscription<void> watchIds({String fileExtension = 'jpg'}) {
+  StreamSubscription<void> watchIds({String? fileExtension}) {
     return db.watch('''
       SELECT photo_id FROM users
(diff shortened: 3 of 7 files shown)


stevensJourney (commented):
Thanks for the work done here. I recognise some of the current limitations the Dart Attachments helper has. This is an improvement.

I recently worked on the [Kotlin](https://github.com/powersync-ja/powersync-kotlin/pull/159) and [Swift](https://github.com/powersync-ja/powersync-swift/pull/35) Attachment helpers. I also noticed this inconvenience during development and opted for a different approach which might align better with your thought of 

> I'm wondering if it shouldn't be pushed further to be able to specify a Map<String, String?> for <filename, extension?>

The Kotlin and Swift implementations allow the watched query to return more than just the attachment ID. They allow returning an array of "partial attachments" which can include a filename and fileExtension. These values are resolved if not provided.

I personally am in favour of updating the watched query result to contain a partial Attachment record type which can contain more information.

simolus3 (commented):
First of all, apologies for the delay in reviewing this. We are also working on a rewrite of the attachments helper package (https://github.com/powersync-ja/powersync.dart/pull/311) based on lessons learned writing similar APIs for our Kotlin and Swift SDKs.

That doesn't mean these changes aren't valuable, and we can still merge and release this since it's an immediate improvement. But I want to point out that it would likely get replaced in the near future.

n0mad-d3v:
Hey @simolus3, 
First of all, thanks a lot for the review 🙏

Regarding this PR, I'll do the changes anyway, mainly from a learning perspective 🤓
But no worries if it doesn't get merged or if the new version you're working on guys is better 🤗 I'm glad this is reworked on your side and really am impatient to test it! 
I'll update this as soon as I can and leave you the choice to merge or not 👍

n0mad-d3v:
I think that you're progressing much faster than me on the #311 😅
Sorry, I had lot of work and didn't manage to progress on this one. I'll close it if that's fine for you 🤓

PR #273, opened 2025-04-28
Add new schema options

simolus3:
This exposes the new table options introduced in version 0.3.13 of the core extension, namely:

- `include_old` (either a boolean or an array of column names): Adds an `old` entry to the `ps_crud.data` JSON object that includes old values for updates.
 - `include_old_only_when_changed`: Skip unchanged values from updates.
- `include_metadata`: Adds a `_metadata` column that can be written to for updates. It's reported through `CrudEntry.metadata` afterwards.
- `ignore_empty_update`: Don't create `ps_crud` entries for `UPDATE` statements that didn't change any values. 

Most of this feature is tested in the core extension, this adds a few simple tests ensuring we forward options correctly.

### DIFF
--- a/packages/powersync_core/lib/src/crud.dart
+++ b/packages/powersync_core/lib/src/crud.dart
@@ -73,7 +73,7 @@ class CrudEntry {
   /// An optional metadata string attached to this entry at the time the write
   /// has been issued.
   ///
-  /// For tables where [Table.includeMetadata] is enabled, a hidden `_metadata`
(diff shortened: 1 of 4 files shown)


stevensJourney (commented):
Changes here LGTM. I think this is good-to-go after we finalize the naming conventions of these new additions.

stevensJourney (approved):
Looks good to me

PR #281, opened 2025-05-15
Refactor sync client and options

simolus3:
This merges all options affecting the sync client (retry duration, crud throttle and client parameters) into a single object (`SyncOptions`) that can be passed to `connect()`. The old field for the retry duration and the independent parameters for throttling and client params have been deprecated in favor of those options.
The main motivation for the options class is that we'll have additional options in the future (e.g. to control whether to use the Dart or the Rust sync client, or whether to use websockets vs. HTTP streams). Having these options all in one place makes it easier to expand them later.

This also applies some refactoring to the client, with a focus on how we handle aborting syncs and errors:

1. Instead of passing individual callbacks to the `StreamingSyncImplementation`, connector methods are now grouped with an interface. The old callbacks didn't have the same name as the methods on the connector, which was confusing to me (e.g. `invalidCredentialsCallback` should call `prefetchCredentials` - not exactly obvious).
2. Fix `addBroadcast` to also cancel all streams when one of them emits a `done` event. Previously, we might leak subscriptions.
3. The sync client had a "local ping" controller emitting `null` to advance the sync iteration. I've introduced sealed classes to be more precise about the event advancing the iteration.
4. This simplifies closing the HTTP client and active streams. Previously, we would close the client to abort the current connection, and there was a comment there mentioning that "there is no other known way to cancel open streams". I'm not sure if I'm overlooking anything, but it seems to me that we can and should just use `StreamSubscription.cancel`? Doing that on the response stream will clean resources, and we can simply close the client afterwards. Thanks to fix number two, we can just close the "local ping" controller to cancel all source streams, which will make `_streamingSyncIteration` break out of the `await for` loop after everything has been cleaned up. So we no longer have to close the HTTP client just to cancel the stream.

### DIFF
--- a/packages/powersync_core/test/in_memory_sync_test.dart
+++ b/packages/powersync_core/test/in_memory_sync_test.dart
@@ -9,6 +9,7 @@ import 'package:powersync_core/src/sync/streaming_sync.dart';
 import 'package:powersync_core/src/sync/protocol.dart';
 import 'package:test/test.dart';
 
+import 'bucket_storage_test.dart';
 import 'server/sync_server/in_memory_sync_server.dart';
 import 'utils/abstract_test_utils.dart';
 import 'utils/in_memory_http.dart';
@@ -652,6 +653,23 @@ void main() {
         await expectLater(status,
             emits(isSyncStatus(downloading: false, downloadProgress: isNull)));
       });
+    });
+
+    test('stopping closes connections', () async {
+      final status = await waitForConnection();
+
+      syncService.addLine({
+        'checkpoint': Checkpoint(
+          lastOpId: '4',
+          writeCheckpoint: null,
+          checksums: [checksum(bucket: 'a', checksum: 0)],
+        )
+      });
+
+      await expectLater(status, emits(isSyncStatus(downloading: true)));
+      await syncClient.abort();
+
+      expect(syncService.controller.hasListener, isFalse);
     });
   });
 }


packages/powersync_core/lib/src/database/powersync_db_mixin.dart
```diff
@@ -269,17 +271,30 @@ mixin PowerSyncDatabaseMixin implements SqliteConnection {
...
+  ///
   /// Status changes are reported on [statusStream].
   Future<void> connect({
     required PowerSyncBackendConnector connector,
-    Duration crudThrottleTime = const Duration(milliseconds: 10),
-    Map<String, dynamic>? params,
+    SyncOptions? options,
+    @Deprecated('Use SyncOptions.crudThrottleTime instead')
+    Duration? crudThrottleTime,
+    @Deprecated('Use SyncOptions.params instead') Map<String, dynamic>? params,
```
rkistner, line 284: I wonder if we should keep this as a separate direct argument here?

I imagine in many applications the params could change dynamically, while the other options would be fairly constant. So to me it makes sense to group all other options (retryDelay, crudThrottleTime, and other future options), but keeping the params separate. We could still group it in the resolved options, to keep the implementation simpler.

rkistner (commented):
Happy with the internal cases.

I can't remember why I didn't go for just cancelling the subscriptions. Just make sure to test the edge cases properly here. I think we cover at least some of it with unit tests, but not sure how thorough the tests are.

rkistner (approved):

PR #290, opened 2025-06-10
WIP: Manual schema management API

davidmartos96:
Here is a proposed API for the manual schema management in the raw tables feature.

A new flag is added to the PowerSyncDatabase factories `manualManagement`, and a new function to mark the schema as ready `db.markSchemaAsReady()`

When using this manual mode, the developer is responsible for calling `db.updateSchema(schema)`. Alternatively we could call automatically the updateSchema function in the `markSchemaAsReady()` function, but we believe it makes sense to be all manual and not mix.

While the schema is not ready, the public functions that are not related to executing SQL are guarded by `_assertSchemaIsReady()`, to prevent calling `connect` for example.

A working example with this is the branch `example-raw-tables` in https://github.com/davidmartos96/powersync.dart with the existing example `supabase-todolist`

Pending changes here are:

- [ ] Docs
- [X] Pass the schema down to `syncIteration()` in the web worker implementation. We are not familiar with the communication protocol it uses.
- [ ] Tests


packages/powersync_core/lib/src/schema.dart
```diff
@@ -315,6 +316,60 @@ class Column {
...
+
+  Map<String, dynamic> toJson() => {
+        'sql': sql,
+        'params': params,
+      };
+}
+
+sealed class PendingStatementValue {
+  dynamic toJson();
+}
```
simolus3, line 352: Can we make the subclasses private and expose them as factories only? E.g.

```dart
sealed class PendingStatementValue {
  factory PendingStatementValue.id() = _PendingStmtValueId;
  factory PendingStatementValue.column(String column) = _PendingStmtValueColumn;

  dynamic toJson();
}
```

packages/powersync_core/lib/src/database/powersync_db_mixin.dart
```diff
@@ -110,10 +114,36 @@ mixin PowerSyncDatabaseMixin implements SqliteConnection {
     statusStream = statusStreamController.stream;
     updates = powerSyncUpdateNotifications(database.updates);
 
+    _manualSchemaManagementCompleted = false;
```
simolus3, line 117: I'm not super happy with this design, but I don't see a better option either.

One thing that looks cleaner might be to have users open their own databases (from `sqlite_async`), initialize the schema there and then pass the database to `PowerSyncDatabase.withDatabase`. But that would make the database tricky to use with wrappers like drift because we can't swap out the earlier database with the other one later.
So this might be the best option we have here, but I'll ping @rkistner for API design advice as well.

davidmartos96:
One thing we noticed while doing this PR is the `logger` param in `PowerSyncDatabase.withDatabase` actually says `loggers`. Technically it's a breaking change, but maybe you guys want to fix it in a separate commit

PR #295, opened 2025-07-02
Fix http error parsing

rkistner:
We should catch Exception, not Error. This resulted in errors like this:

```
FormatException: Unexpected character (at character 1)
<html>
^
```

With this change, we're getting the expected exception:
```
SyncResponseException: 503 Service Unavailable
```

### DIFF
--- a/packages/powersync_core/lib/src/exceptions.dart
+++ b/packages/powersync_core/lib/src/exceptions.dart
@@ -38,15 +38,11 @@ class SyncResponseException implements Exception {
       http.StreamedResponse response) async {
     try {
       final body = await response.stream.bytesToString();
-      final decoded = convert.jsonDecode(body);
-      final details = _stringOrFirst(decoded['error']?['details']) ?? body;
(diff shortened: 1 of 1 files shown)


packages/powersync_core/lib/src/exceptions.dart
```diff
@@ -58,12 +58,7 @@ class SyncResponseException implements Exception {
       final details = _stringOrFirst(decoded['error']?['details']) ?? body;
       final message = '${response.reasonPhrase ?? "Request failed"}: $details';
       return SyncResponseException(response.statusCode, message);
-    } on FormatException catch (_) {
```
simolus3, line 61: This format exception looks correct to me, but I'm fine catching all `Exception`s as well.

I am wondering if we should refactor the common part of this into a say `static SyncResponseException fromResponseBody(String body)` though?

simolus3 (dismissed):
One nit about merging the shared part of the two methods into one since they're identical, but apart from that this looks good to me 👍

simolus3 (approved):

PR #298, opened 2025-07-07
Fix typo in method parameter: loggers -> logger

davidmartos96:
`PowerSyncDatabase.withDatabase` was declaring `loggers` instead of `logger`. This changes it while keeping the wrong one as deprecated to avoid breaking.


simolus3 (approved):
Thanks! I think using `logger` consistently is the right thing to do here 👍

PR #311, opened 2025-07-22
Attachment package refactor

dean-journeyapps:
This ports the new attachment APIs from the Kotlin and Swift SDKs over to Dart. For simplicity, the new APIs are added to `package:powersync_core` (with an independent `package:powersync_core/attachments/attachments.dart` import).

Unlike the old attachments helper package, the new APIs generally support the web platform as well (instead of causing compilation errors). However, one needs to explicitly provide a storage implementation, and the only builtin implementation (`IOLocalStorage`) will not work on the web. We can revisit proper web support at a later point in time.


packages/powersync_attachments_stream/pubspec.yaml
```diff
@@ -0,0 +1,32 @@
+name: powersync_attachments_stream
+description: "A helper library for handling attachments when using PowerSync."
+version: 0.0.1
+repository: https://github.com/powersync-ja/powersync.dart
+homepage: https://www.powersync.com/
+
+environment:
+  sdk: ^3.8.1
+  flutter: ">=1.17.0"
```
simolus3, line 9: I think this package shouldn't have to depend on Flutter, right? For those packages you can remove flutter dependencies and the `platforms` section since it will work on everything.

packages/powersync_attachments_stream/lib/src/utils/mutex.dart
```diff
@@ -0,0 +1,33 @@
+// lib/src/utils/mutex.dart
+
+import 'dart:async';
+
+class AsyncMutex {
```
simolus3, line 5: I think there's a mutex implementation exposed by `package:sqlite_async` which we could re-use here.

packages/powersync_attachments_stream/lib/src/local_storage.dart
```diff
@@ -0,0 +1,84 @@
...
+  Future<void> delete(String id);
+  
+  /// Retrieves metadata associated with stored data
+  /// 
+  /// [id] - Unique identifier of the data
+  /// Returns metadata map if found, null otherwise
+  Future<Map<String, dynamic>?> getMetadata(String id);
+}
+
+class IOLocalStorage implements LocalStorage {
```
simolus3, line 32: I think it would be nice if we could move code that depends on `dart:io` into different files.

Then, on the public level (outside of `src/`), we could have an import for users who want to use this on the web (writing their own local storage implementations for e.g. the web filesystem APIs, but being unable to import `dart:io`).
So there could be e.g. a `package:powersync_attachments_stream/powersync_attachments_stream.dart` for the default things that depend on `dart:io`, and another `package:powersync_attachments_stream/common.dart` for the subset of files that work on all platforms.

packages/powersync_attachments_stream/lib/src/local_storage.dart
```diff
@@ -0,0 +1,84 @@
...
+  IOLocalStorage(this.baseDir);
+
+  File _fileFor(String id) => File(p.join(baseDir.path, id));
+  File _metaFileFor(String id) => File(p.join(baseDir.path, '$id.meta.json'));
+
+  @override
+  Future<void> save(String id, List<int> bytes, {String? mediaType, Map<String, dynamic>? metadata}) async {
+    await baseDir.create(recursive: true);
+    await _fileFor(id).writeAsBytes(bytes);
+    if (mediaType != null || metadata != null) {
```
simolus3, line 44: I'm not sure about saving the media type like this, what's the motivation for not storing it in the attachments table instead?

packages/powersync_attachments_stream/lib/src/attachment_service.dart
```diff
@@ -0,0 +1,13 @@
+import 'attachment_state.dart';
+import 'attachment_context.dart';
+
+abstract class AttachmentService {
+  Future<void> init();
+  Future<void> close();
+
+  Stream<List<Attachment>> watchActiveAttachments();
```
simolus3, line 8: Do you think it makes sense to move of the methods related to fetching attachments to the `AttachmentContext` class like it's done in Kotlin?

packages/powersync_attachments_stream/lib/src/attachment_queue_service.dart
```diff
@@ -1,66 +1,370 @@
...
+  final Stream<List<WatchedAttachmentItem>> Function() watchAttachments;
   final LocalStorage localStorage;
-  final SyncErrorHandler errorHandler;
-  final AsyncMutex _mutex = AsyncMutex();
+  final String attachmentsQueueTableName;
+  final SyncErrorHandler? errorHandler;
+  final Duration syncInterval;
+  final int archivedCacheLimit;
+  final Duration syncThrottleDuration;
+  final List<String>? subdirectories;
```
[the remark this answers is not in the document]
dean-journeyapps: Local storage adapter creates new directories on demand now. Also add tests for that.

packages/powersync_attachments_stream/lib/src/attachment_queue_service.dart
```diff
@@ -0,0 +1,404 @@
+// Implements the attachment queue for PowerSync attachments.
+//
+// This class manages the lifecycle of attachment records, including watching for new attachments,
+// syncing with remote storage, handling uploads, downloads, and deletes, and managing local storage.
+// It provides hooks for error handling, cache management, and custom filename resolution.
+
+import 'dart:async';
+import 'dart:typed_data';
+import 'package:logging/logging.dart';
+import './storage/io_local_storage.dart';
```
simolus3, line 10: It would be nice if `AttachmentQueue` were available from `common.dart` and the web, and this import is the only thing preventing that.

It looks like it's only used as a default for `localStorage`. I see it's helpful to not make that a required argument, but I wonder if in this case it may be better to force users to spell out `IOLocalStorage()` when constructing `AttachmentQueue`s to be explicit about the `dart:io` dependency.

I think my preferred API for Dart would be to move the `attachmentsDirectory` parameter into the `IOLocalStorage()` class, I've started a discussion on this offline.

packages/powersync_attachments_stream/lib/src/implementations/attachment_context.dart
```diff
@@ -0,0 +1,161 @@
...
+
+    log.info('Deleting ${archivedAttachments.length} archived attachments (exceeding maxArchivedCount=$maxArchivedCount)...');
+    // Call the callback with the list of archived attachments before deletion
+    await callback(archivedAttachments);
+
+    // Delete the archived attachments from the table
+    final ids = archivedAttachments.map((a) => a.id).toList();
+    if (ids.isNotEmpty) {
+      final placeholders = List.filled(ids.length, '?').join(',');
+      await db.execute('DELETE FROM $table WHERE id IN ($placeholders)', ids);
```
simolus3, line 128: Even with the 1000 item limit this can lead to very large statements, I think especially some older SQLite versions also have a limit on prepared statement parameters that may be lower than that. Let's use a statement with a single id and run that multiple times:

```suggestion
      await db.executeBatch('DELETE FROM $table WHERE id = ?', [
        for (final id in ids) [id],
      ]);
```

packages/powersync_attachments_stream/lib/src/implementations/attachment_service.dart
```diff
@@ -0,0 +1,63 @@
...
+
+    yield* stream;
+  }
+
+  @override
+  Future<T> withContext<T>(Future<T> Function(AbstractAttachmentContext ctx) action) {
+    // Simple mutex using chained futures
+    final completer = Completer<T>();
+    _mutex = _mutex.then((_) => action(_context)).then(completer.complete).catchError(completer.completeError);
+    return completer.future;
```
simolus3, line 61: Let's also use a `package:sqlite_async` `Mutex` for this here

packages/powersync_attachments_stream/lib/src/storage/io_local_storage.dart
```diff
@@ -0,0 +1,127 @@
...
+/// Handles file and directory operations for attachments.
+class IOLocalStorage implements AbstractLocalStorageAdapter {
+  final Directory baseDir;
+
+  IOLocalStorage([Directory? baseDir])
+    : baseDir = baseDir ?? Directory.systemTemp;
+
+  File _fileFor(String filePath) => File(p.join(baseDir.path, filePath));
+  File _metaFileFor(String filePath) =>
+      File(p.join(baseDir.path, '$filePath.meta.json'));
```
simolus3, line 33: It looks like we're not using `_metaFileFor` anymore outside of copying and deleting? So it seems like we can remove this entirely?

packages/powersync_attachments_stream/lib/src/storage/io_local_storage.dart
```diff
@@ -0,0 +1,127 @@
...
+/// - Delete files and their metadata
+/// - Copy files and their metadata
+/// - Create and remove directories (creates all necessary parents dynamically)
+/// - Check file existence
+
+import 'dart:async';
+import 'dart:io';
+import 'dart:typed_data';
+import 'package:path/path.dart' as p;
+import '../abstractions/local_storage.dart';
```
simolus3, line 20: nit: Typically one inserts an empty line between `dart:` and `package:` imports and then another one before local imports. All the [effective Dart samples](https://dart.dev/effective-dart/style#do-place-dart-imports-before-other-imports) do that, but it's not spelled out explicitly. I've added a suggestion for this here, it affects other files as well.

```suggestion
import 'dart:async';
import 'dart:io';
import 'dart:typed_data';

import 'package:path/path.dart' as p;

import '../abstractions/local_storage.dart';
```

packages/powersync_attachments_stream/lib/src/attachment_queue_service.dart
```diff
@@ -0,0 +1,404 @@
...
+        if (notInWatchedItems) {
+          switch (attachment.state) {
+            case AttachmentState.queuedDelete:
+            case AttachmentState.queuedUpload:
+              if (attachment.hasSynced) {
+                attachmentUpdates.add(
+                  attachment.copyWith(state: AttachmentState.archived),
+                );
+              }
+              break;
```
simolus3, line 275: You don't need this `break` anymore in recent Dart versions.

packages/powersync_attachments_stream/lib/src/storage/io_local_storage.dart
```diff
@@ -0,0 +1,127 @@
...
+/// the Dart IO library. It is used as the local storage backend for attachments
+/// in the PowerSync attachments system.
+///
+/// Features:
+/// - Save files from streams (creates directories and all necessary parents dynamically if they do not exist)
+/// - Read files as streams
+/// - Delete files and their metadata
+/// - Copy files and their metadata
+/// - Create and remove directories (creates all necessary parents dynamically)
+/// - Check file existence
```
simolus3, line 14: This documentation comment is dangling, it doesn't actually document anything. You can add a `library;` directive after it to attach the comment to this file explicitly, but IMO internal libraries like this one don't need documentation comments.

You could make it a regular comment (`//`), or extract relevant bits to the documentation of the class and then remove this comment entirely. This also applies to some other files (except those in `lib/` outside of `src/` - those should have a `library` directive and a documentation comment on the library).

packages/powersync_attachments_stream/lib/src/sync/syncing_service.dart
```diff
@@ -0,0 +1,324 @@
...
+      triggerSync();
+    });
+  }
+
+  StreamTransformer<T, T> _throttleTransformer<T>(Duration throttle) {
+    return StreamTransformer.fromHandlers(
+      handleData: (data, sink) {
+        sink.add(data);
+        // Simple throttle implementation - just delay the next event
+        Future.delayed(throttle);
```
simolus3, line 113: This future is unawaited and will not actually throttle anything.

I think it would be easier to not throttle the stream here and instead add a duration parameter to `watchActiveAttachments`. We can forward that to `db.watch()` instead of duplicating the throttle implementation here. That means we wouldn't be throttling `manualTriggers` anymore, but I think that's fine.

packages/powersync_attachments_stream/lib/src/attachment_queue_service.dart
```diff
@@ -0,0 +1,404 @@
...
+      }
+
+      await context.saveAttachments(attachmentUpdates);
+    });
+  }
+
+  /// Creates a new attachment locally and queues it for upload.
+  /// The filename is resolved using [resolveNewAttachmentFilename].
+  Future<Attachment> saveFile({
+    required Stream<Uint8List> data,
```
simolus3, line 291: I noticed we had to cast in the demo to get a `Stream<Uint8List>`, so I wonder if we should take a `List<int>` as an input here instead (when we expose assets as bytes we should definitely use `Stream<Uint8List>` for that, but it looks like for local writes we can get away with `List<int>` too).

simolus3 (commented):
A few more comments, most of them pretty minor again :)

packages/powersync_attachments_stream/lib/powersync_attachments_stream.dart
```diff
@@ -0,0 +1,4 @@
+/// Default exports for native platforms (dart:io). For web, use 'common.dart'.
+export 'common.dart';
+export 'src/storage/io_local_storage.dart';
+export 'src/attachment_queue_service.dart';
```
simolus3, line 4: This is exported by `common.dart` now and can be removed here.

packages/powersync_attachments_stream/lib/src/abstractions/attachment_context.dart
```diff
@@ -0,0 +1,49 @@
...
+/// Context for performing Attachment operations.
+///
+/// This is typically provided through a locking/exclusivity method and allows
+/// safe, transactional operations on the attachment queue.
+abstract class AbstractAttachmentContext {
+  /// Delete the attachment from the attachment queue.
+  ///
+  /// [id]: The ID of the attachment to delete.
+  /// [tx]: The database context to use for the operation.
+  Future<void> deleteAttachment(String id, dynamic context);
```
simolus3, line 12: This shouldn't be dynamic, can this be a `SqliteWriteContext`?

simolus3, line 12: Also this function isn't called at all, I assume that should happen in `SyncingService.deleteAttachment`?

packages/powersync_attachments_stream/lib/src/abstractions/attachment_context.dart
```diff
@@ -0,0 +1,49 @@
...
+  ///
+  /// Returns true if all items have been deleted. Returns false if there might be more archived items remaining.
+  Future<bool> deleteArchivedAttachments(Future<void> Function(List<Attachment>) callback);
+
+  /// Upserts an attachment record given a database connection context.
+  ///
+  /// [attachment]: The attachment to upsert.
+  /// [context]: The database transaction/context to use for the operation.
+  /// Returns the upserted [Attachment].
+  Future<Attachment> upsertAttachment(Attachment attachment, dynamic context);
```
simolus3, line 48: `context` should likely be a `SqliteWriteContext` here as well?

packages/powersync_attachments_stream/lib/src/abstractions/attachment_context.dart
```diff
@@ -0,0 +1,49 @@
+import '../attachment.dart';
+
+/// Context for performing Attachment operations.
+///
+/// This is typically provided through a locking/exclusivity method and allows
+/// safe, transactional operations on the attachment queue.
+abstract class AbstractAttachmentContext {
```
simolus3, line 7: I don't really understand the purpose of making this and `AbstractAttachmentService` an abstract class:

1. It doesn't look like it makes sense for users to write their own implemenation of this, so by exposing this interface we can't change or add methods without a breaking change.
2. It means that users have to manually create `AttachmentServiceImpl`, it would be nicer if the implementation class was just named `AttachmentService` directly and we didn't have a separate interface.

packages/powersync_attachments_stream/lib/src/abstractions/remote_storage.dart
```diff
@@ -0,0 +1,26 @@
+import 'dart:async';
+ 
+import '../attachment.dart';
+
+/// Adapter for interfacing with remote attachment storage.
+abstract class AbstractRemoteStorageAdapter {
```
simolus3, line 6: Let's make this an interface class and just name it `RemoteStorageAdapter`:

```suggestion
abstract interface class RemoteStorageAdapter {
```

packages/powersync_attachments_stream/lib/src/abstractions/sync_error_handler.dart
```diff
@@ -0,0 +1,36 @@
+import '../attachment.dart';
+
+/// Interface for handling errors during attachment operations.
+/// Implementations determine whether failed operations should be retried.
+/// Attachment records are archived if an operation fails and should not be retried.
+abstract class AbstractSyncErrorHandler {
```
simolus3, line 6: Let's make this an interface class and just name it `SyncErrorHandler`:

```suggestion
abstract interface class SyncErrorHandler {
```

packages/powersync_attachments_stream/lib/src/abstractions/local_storage.dart
```diff
@@ -0,0 +1,34 @@
+import 'dart:typed_data';
+
+abstract class AbstractLocalStorageAdapter {
```
simolus3, line 3: Let's make this an interface class and just name it `LocalStorageAdapter`:

```suggestion
abstract interface class LocalStorageAdapter {
```

packages/powersync_attachments_stream/lib/src/attachment_queue_service.dart
```diff
@@ -0,0 +1,392 @@
...
+
+  /// Resolves the filename for new attachment items.
+  /// Concatenates the attachment ID and extension by default.
+  Future<String> resolveNewAttachmentFilename(
+    String attachmentId,
+    String? fileExtension,
+  ) async {
+    return fileExtension != null
+        ? '$attachmentId.$fileExtension'
+        : '$attachmentId.dat';
```
simolus3, line 193:
```suggestion
    return '$attachmentId.${fileExtension ?? 'dat'}';
```

packages/powersync_attachments_stream/lib/src/attachment_queue_service.dart
```diff
@@ -0,0 +1,392 @@
...
+      final currentAttachments = await context.getAttachments();
+      final List<Attachment> attachmentUpdates = [];
+
+      for (final item in items) {
+        final existingQueueItem = currentAttachments
+            .cast<Attachment?>()
+            .firstWhere(
+              (a) => a != null && a.id == item.id,
+              orElse: () => null,
+            );
```
simolus3, line 212: I think a more intuitive way to express this without the `cast()` may be
```suggestion
            .where((a) => a.id == item.id)
            .firstOrNull;
```

packages/powersync_attachments_stream/lib/src/attachment_queue_service.dart
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  /// Creates a new attachment locally and queues it for upload.
+  /// The filename is resolved using [resolveNewAttachmentFilename].
+  Future<Attachment> saveFile({
+    required List<int> data,
+    required String mediaType,
+    String? fileExtension,
+    String? metaData,
+    required Future<void> Function(dynamic context, Attachment attachment)
```
simolus3, line 289:
```suggestion
    required Future<void> Function(SqliteWriteContext context, Attachment attachment)
```

packages/powersync_attachments_stream/lib/src/attachment_queue_service.dart
```diff
@@ -0,0 +1,392 @@
...
+        return await attachmentContext.upsertAttachment(attachment, tx);
+      });
+    });
+  }
+
+  /// Queues an attachment for delete.
+  /// The default implementation assumes the attachment record already exists locally.
+  Future<Attachment> deleteFile({
+    required String attachmentId,
+    required Future<void> Function(dynamic context, Attachment attachment)
```
simolus3, line 326:
```suggestion
    required Future<void> Function(SqliteWriteContext context, Attachment attachment)
```

stevensJourney (dismissed):
I left some minor comments. Overall I'm very happy with the changes here. The code is well documented and very clean :)

packages/powersync_core/lib/src/attachments/attachment_queue_service.dart
```diff
@@ -0,0 +1,443 @@
...
+    final List<Attachment> updates = [];
+
+    for (final attachment in attachments) {
+      // Only check attachments that should have local files
+      if (attachment.localUri == null) {
+        // Skip attachments that don't have localUri (like queued downloads)
+        continue;
+      }
+
+      final exists = await _localStorage.fileExists(attachment.localUri!);
```
stevensJourney, line 431: It's only an iOS related issue, so slightly less relevant here, but since we are working on this package now. It could be good to implement the changes from here in this package https://github.com/powersync-ja/powersync-swift/pull/74.

stevensJourney (approved):
This looks good to me 👍 . I'm happy to release this as is if needed, we can port the localURI changes later - potentially when we update the Kotlin SDK.

stevensJourney (approved):
The changes here look amazing to me.
The last thing I can think of is we might want to update the current Attachments package to maybe deprecate that and add a small reference to this new functionality.

simolus3:
>  The last thing I can think of is we might want to update the current Attachments package to maybe deprecate that and add a small reference to this new functionality.

I will do that in a separate PR once we prepare the release 👍

PR #317, opened 2025-08-13
Sync streams

simolus3:
This adds support for sync streams to the Dart SDK by implementing the new methods discussed in the "2025-03 Sync streams" design doc ([public excerpt](https://github.com/powersync-ja/powersync-service/discussions/347)). Namely: 

- `db.syncStream(name, [params])` returns a `SyncStream` instance, which can be used to `subscribe()` to that stream.
- calling `subscribe()` on a `SyncStream` returns a `SyncStreamSubscription`, giving you access to `waitForFirstSync()` and `unsubscribe()` on that subscription.
- the `SyncStatus` interface includes a list of `SyncSubscriptionDefinition`s, describing all streams that the client is subscribed to (either due to an explicit subscriptions or because the stream has `auto_subscribe: true`). That interface also reports per-stream download progress.

Additional implementation changes support this:

1. We want a `hasSynced` and `lastSyncedAt` for individual streams. The core extension gained a `powersync_offline_sync_status()` function to help with this, meaning that SDKs no longer have to query the underlying tables directly.
2. In the app itself, active subscriptions (that currently have a `SyncStreamSubscription` instance attached to them) can change often and quickly. The TTL feature ensures we don't have to recreate a streaming sync socket every time they change. This is mostly implemented in the sync service, we just send a `subscriptions_updated` event and the Rust client tells us whether to restart the iteration or not.
3. I have moved some connection and subscription management logic into the `ConnectionManager` class to not complicate the mixin even further.

I have adopted streams in the `supabase-todolist` demo, with the following sync rules:

```
streams:
  lists:
    query: SELECT * FROM lists
    auto_subscribe: true
  todos:
    query: SELECT * FROM todos WHERE list_id = subscription.parameter('list')
```

The idea is that lists are shown by default, and entries in a list are loaded and cached when the page is opened for the first time. This works well, but I really don't like that the API currently requires [50 lines of user code for that](https://github.com/powersync-ja/powersync.dart/blob/d4f676b42c172838f19ca801074a8c4f93bbff33/demos/supabase-todolist/lib/widgets/todo_list_page.dart#L54-L102). We should probably look into a `powersync_flutter` package with utilities to simplify this eventually.

### DIFF
--- a/packages/powersync_core/test/utils/abstract_test_utils.dart
+++ b/packages/powersync_core/test/utils/abstract_test_utils.dart
@@ -83,7 +83,7 @@ abstract mixin class TestPowerSyncFactory implements PowerSyncOpenFactory {
       database: SqliteDatabase.singleConnection(
           SqliteConnection.synchronousWrapper(raw)),
       logger: logger ?? Logger.detached('PowerSync.test'),
-      schema: schema,
+      schema: customSchema ?? schema,
     );
   }
 }


stevensJourney (approved):
Happy with this from my side :D

stevensJourney (approved):
🚀

PR #325, opened 2025-09-01
Add SPM support to `powersync_flutter_libs`

simolus3:
For native dependencies on Apple platforms, Flutter is [migrating to Swift package manager](https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-app-developers) to replace CocoaPods.

SPM support is experimental at the moment (users have to opt-in with `flutter config --enable-swift-package-manager`). But since we can support both in `powersync_flutter_libs`, we probably should:

- We're still early, but CocoaPods trunk will become [read-only](https://blog.cocoapods.org/CocoaPods-Specs-Repo/) in 15 months. So we have to migrate all our SDKs eventually.
- Users have a similar interest, and they can only switch to SPM after all their dependencies have. So this unblocks users.

So, this PR:

- merges `ios` and `macos` podspecs for `powersync_flutter_libs` into a single `darwin` directory for simplicity.
- creates a new SPM package for `powersync_flutter_libs` depending on our [core extension dependency](https://github.com/powersync-ja/powersync-sqlite-core-swift/tree/main).
- updates `demos/supabase-todolist-drift` to use SPM exclusively.
- updates `demos/supabase-todolist` to use CocoaPods (and explicitly opt out of SPM) so that we can keep testing both variants of `powersync_flutter_libs`.

I ran both updated projects on macOS and an iOS simulator for verification.


simolus3:
> I see some of the platform support versions have changed - are these differences intentional?

I think we could keep referencing the older versions as well, but iOS 13 and macOS 10.15 are the minimum versions [supported by Flutter](https://docs.flutter.dev/reference/supported-platforms). I agree that we should make these consistent though.

PR #388, opened 2026-03-30
Migrate to pub workspace

simolus3:
This migrates this repository to be a [pub workspace](https://dart.dev/tools/pub/workspaces). Previously, we've used melos to call `pub get` recursively and install `dependency_overrides` between packages. Pub workspaces can do the same, while also being much more efficient and ensuring that we have consistent versions in the entire repository (there's a single `pubspec.lock` only).

I had to fix some scripts (`downloadWebAssets` used hardcoded `package_config.json` paths that don't work with workspaces) and used this opportunity to update a few dependencies in demos. For the drift todolist demo, some changes were necessary to apply a major `riverpod` update. Demos are now covered by `dart analyze` check as well.

For now, we keep using melos with the pub workspace. After adopting build hooks (https://github.com/powersync-ja/powersync.dart/pull/386) and upgrading to the latest version of `sqlite3`, we'll likely only have a single SDK package and won't need melos tools anymore.


.github/workflows/check.yml
```diff
@@ -19,43 +19,24 @@ jobs:
...
+        run: |
+          find . -name "app_config_template.dart" -exec sh -c 'cp "$1" "$(dirname "$1")/app_config.dart"' _ {} \;
+          touch demos/supabase-trello/.env
       - name: Lint
-        run: melos analyze:packages
+        run: dart analyze --fatal-infos --fatal-warnings
       - name: Publish dry-run
         run: melos publish --dry-run --yes
 
-  pana:
```
simolus3, line 28: I'll add the pana score check back after the other items planned for the Dart SDK. It would likely break with the intermediate state as we merge packages.

rkistner (approved):

PR #394, opened 2026-04-07
Merge web workers

simolus3:
Currently, the Dart SDK relies on two web workers: One hosting the database (can either be a shared or dedicated worker), and another one coordinating the sync process across tabs (always a shared worker). With new `sqlite3_web` APIs, we can merge them into one. The main benefit here is a slightly easier setup for users since there is just a single file to manage. 

It also reduces app size a bit, but the workers have already been fairly small so it's not a huge deal:

- Latest release (1.18.0): 676KiB in total for both workers.
- Current state on `dev` with some size optimizations from `sqlite3_web`: 635 KiB in total for both workers.
- With this PR: A single worker, 453KiB in size.

The basic idea here is to encapsulate messages sent to shared workers via `SharedWorkerMessage`, which describes whether a message is intended for the database or sync part of the worker. By providing a custom `WorkerConnector`, we instruct the `sqlite3_web` package to use that structure for its messages as well.


rkistner (approved):
Just to confirm my understanding: My first thought was "why not use separate ports for db vs sync worker?", but it looks like this is effectively what's being done here? As in, the only case the shared message is used for the sync worker is to setup that message port?

simolus3:
> As in, the only case the shared message is used for the sync worker is to setup that message port?

Yes. For the sync worker we'll only ever send a single port as a top-level message. `sqlite3_web` happens to do the same (it actually sends a lock name as well to detect when the client has closed). So your intuition is correct, in practice each tab would send exactly two messages to the top-level shared worker port (one for sync and one for `sqlite3_web`). The rest flows though message channels.

(we could make this explicit and implement the shared worker wrapper for `sqlite3_web` through another pair of ports. But that's mostly wasteful, as `sqlite3_web` also uses a `MessageChannel` internally to establish communication.)

PR #395, opened 2026-04-08
Remove Dart client implementation

simolus3:
The Rust client has been available for around a year at this point, and we've made it the default around four months ago. I'm not aware of any regressions reported on the Dart SDK and we're about to release a major update, so this is a good opportunity to remove the Dart client.

The Dart client doesn't support Sync Streams or raw tables, removing it also means we have less code to maintain and apps using PowerSync can be slightly smaller.


rkistner (approved):
Always great when we can delete code

PR #405, opened 2026-04-14
Don't copy worker file to itself

simolus3:
To compile the web worker, we run `dart compile js -o packages/powersync/assets/powersync_db.worker.js`. Afterwards, we used to copy the emitted JavaScript file to its own path. This is harmless on macOS, but causes the resulting `powersync_db.worker.js` file to be empty on Linux. That in turn caused [release automation failures](https://github.com/powersync-ja/powersync.dart/actions/runs/24385866742/job/71219494021), I had to manually upload the worker file for the last releases.

This PR should fix the release process and ensure we run tests with workers available.

### DIFF
--- a/tool/compile_webworker.dart
+++ b/tool/compile_webworker.dart
@@ -32,13 +32,6 @@ Future<void> main() async {
         'Could not compile db worker.\nstdout: ${dbWorkerProcess.stdout.toString()}\nstderr: ${dbWorkerProcess.stderr.toString()}');
   }
 
-  final workerFile = File(dbWorkerOutputPath);
-
-  // Copy workers to powersync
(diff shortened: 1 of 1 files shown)


LucDeCaf (approved):
Out of curiosity, why _were_ we copying the worker file to itself if it's (seemingly) unnecessary? Just curious :)

simolus3:
> Out of curiosity, why _were_ we copying the worker file to itself if it's (seemingly) unnecessary? Just curious :)

Being to lazy to look it up right now, I think it's probably because we used to emit the worker file in another directory first. I probably changed that at some point of the refactoring and forgot to look at the second half of the file.

PR #406, opened 2026-04-17
Add DevTools extension

simolus3:
Flutter and Dart [DevTools](https://docs.flutter.dev/tools/devtools) are a collection of debuggers and inspectors for Dart apps. Packages can contribute additional tools, and I think this could be interesting for PowerSync! This adds an extension to the `powersync` package. It currently has two pages, which provide these features:

## Overview

grafik

This page shows the full path of the database (which may be convenient when `path_provider` is used and one wants to inspect a local database file), potential sync issues (a warning is shown for pending `ps_crud` rows and when the `$local` bucket is waiting for a checkpoint). This also exports the current token and endpoint passed to PowerSync, these values can be inspected and a deeplink can be used to open the same session in the full diagnostics app.

## SQL console

grafik

This page is also inspired from the diagnostics app, it can be used to run queries against the live database in the app. There are buttons to help select from all tables in the schema (including raw tables).

## How it works

Dart code can provide RPC entrypoints to invoke functions through the VM service protocol used by DevTools. Outside of release builds we do that to provide commands listing all opened PowerSync instances, to run queries or to post notifications on table and sync status changes.

In the DevTools extension, we use these RPC entrypoints to obtain a remote database instance we can use to run queries and to observe the sync status. This allows building the UI pages almost as if the PowerSync database was active in the extension itself.

There's a lot more potential here: We could also show buckets with download information through the new diagnostics protocol of the core extension. This is about as far as I could get in a hackday though, and I don't think it makese sense to invest much more time into it. We don't need this to overlap with the diagnostics app either, it's good enough to show obvious issues and can deeplink to diagnostics for anything more complex than that.


.github/workflows/publish.yml
```diff
@@ -36,12 +36,29 @@ jobs:
...
+        with:
+          flutter-version: "3.x"
+          channel: "stable"
+      - run: dart pub get
+      - run: tool/build.sh
+        name: Build and copy devtools extension
+        working-directory: packages/powersync_devtools_extension
+      
+      - name: Publish to pub.dev
+        run: dart pub lish -f
```
stevensJourney: Looks like this might be a typo? Or a very interesting subcommand for `dart pub` :D ?

stevensJourney (dismissed):
This looks really cool!
I added the `Product Visibility` label to this PR - since I assume we might want to document this feature in the docs.

.github/workflows/publish.yml
simolus3: `lish` is actually the [canonical name for the publish subcommand](https://github.com/dart-lang/pub/blob/b5263b3f581eadbe616e803a89c67772f31e2924/lib/src/command/lish.dart#L32), with `publish` being an alias. But since it obviously causes more confusion, I agree we should use `dart pub publish`.

stevensJourney (approved):

packages/powersync_devtools_extension/lib/ui/overview.dart
```diff
@@ -0,0 +1,370 @@
...
+                        showDialog(
+                          barrierDismissible: true,
+                          context: context,
+                          builder: (_) =>
+                              _TokenDetailsDialog(credentials: credentials),
+                        );
+                      },
+                    ),
+                    ButtonGroupItemData(
+                      label: 'Open in Diagnostics App',
```
benitav, line 146: This should be called Sync Diagnostics Client

PR #410, opened 2026-04-23
Simplify control flow in sync client

### LINKED ISSUE #409: Performance regression in desktop integration tests after upgrading 1.18.0 to 2.0.1
We have an integration test suite running on macOS that previously completed in under 10 seconds with version `1.18.0`. After upgrading to `2.0.1`, the same suite now takes over a minute to run.

I haven't pinpointed the exact cause yet, but it appears synchronization is noticeably slower. Happy to dig deeper and share more details if that would help narrow it down.

On a positive note: in the previous version we had to download libpowersync.dylib manually and wire it up with a custom factory. That's no longer necessary in 2.0.1 ❤️.

simolus3:
This closes https://github.com/powersync-ja/powersync.dart/issues/409. Previously, there was a potential race condition when adding new sync streams because we:

1. First added a `AbortCurrentIteration` local sync event.
2. Which is then handled asynchronously to break out of the client loop once handled, but it's possible for additional server-side events to come in before we get to that.

Instead of handling close events asynchronously, this refactors the control flow of the sync client, allowing us to just break out of a loop when the core extension tells us to instead of having to go through these hoops. We do this by making all instructions that won't interrupt a sync iteration extend a common class, so we can safely express which instructions we can handle in a separate methods and which ones need to be part of the loop.

This also updates the core extension and removes a workaround for a `package:http` bug related to cancelling streams on the web.

### DIFF
--- a/packages/powersync/lib/src/sync/streaming_sync.dart
+++ b/packages/powersync/lib/src/sync/streaming_sync.dart
@@ -141,27 +141,21 @@ class StreamingSyncImplementation implements StreamingSync {
 
   @override
   Future<void> streamingSync() async {
+    assert(_abort == null);
+    final abort = _abort = AbortController();
+
     try {
-      assert(_abort == null);
-      _abort = AbortController();
       clientId = await adapter.getClientId();
       _crudLoop();
-      var invalidCredentials = false;
       while (!aborted) {
-        _state.updateStatus((s) => s.setConnectingIfNotConnected());
         var delayNextIteration = false;
 
         try {
-          if (invalidCredentials) {
-            // This may error. In that case it will be retried again on the next
-            // iteration.
-            await connector.prefetchCredentials();
-            invalidCredentials = false;
-          }
           // Protect sync iterations with exclusivity (if a valid Mutex is provided)
           final (:immediateRestart) = await syncMutex.lock(
-              () => _rustStreamingSyncIteration(),
-              abortTrigger: Future.delayed(_retryDelay));
+            () => _rustStreamingSyncIteration(abort),
+            abortTrigger: Future.delayed(_retryDelay),
+          );
           delayNextIteration = !immediateRestart;
         } catch (e, stacktrace) {
           if (aborted && e is http.ClientException) {
@@ -172,7 +166,6 @@ class StreamingSyncImplementation implements StreamingSync {
           delayNextIteration = true;
           final message = _syncErrorMessage(e);
           logger.warning('Sync error: $message', e, stacktrace);
-          invalidCredentials = true;
 
           _state.updateStatus((s) => s.applyDownloadError(e));
         }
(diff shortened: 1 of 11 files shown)


packages/powersync/lib/src/sync/streaming_sync.dart
```diff
@@ -141,27 +141,21 @@ class StreamingSyncImplementation implements StreamingSync {
 
   @override
   Future<void> streamingSync() async {
+    assert(_abort == null);
+    final abort = _abort = AbortController();
```
simolus3, line 145: This ensures a `_rustStreamingSyncIteration` always has access to the abort controller from its iteration and avoids a bunch of non-null assertions.

packages/powersync/lib/src/sync/streaming_sync.dart
```diff
@@ -141,27 +141,21 @@ class StreamingSyncImplementation implements StreamingSync {
...
+
     try {
-      assert(_abort == null);
-      _abort = AbortController();
       clientId = await adapter.getClientId();
       _crudLoop();
-      var invalidCredentials = false;
       while (!aborted) {
-        _state.updateStatus((s) => s.setConnectingIfNotConnected());
```
simolus3, line 151: The core extension sets `connecting: true` on start, so no need to replicate that.

packages/powersync/lib/src/sync/streaming_sync.dart
```diff
@@ -141,27 +141,21 @@ class StreamingSyncImplementation implements StreamingSync {
...
         var delayNextIteration = false;
 
         try {
-          if (invalidCredentials) {
-            // This may error. In that case it will be retried again on the next
-            // iteration.
-            await connector.prefetchCredentials();
```
simolus3, line 158: I don't know what the point of this is - `invalidCredentials` is set on all exceptions but we also don't actually invalidate credentials? We already drop cached credentials on a 401 error, and obviously we fetch credentials when starting a connection.

But I might be missing something, so a double-check would be appreciated :D

stevensJourney: I also can't really think of a reason for this code existing at this point-in-time (with the Rust sync implementation). I think my best guess is that this might have been used by the legacy Dart streaming sync implementation for some reason.

stevensJourney (approved):
The changes here look good to me :)

packages/powersync/lib/src/sync/streaming_sync.dart
rkistner: The original idea around `prefetchCredentials` in general was that we can trigger the fetch before we actually need it, which avoids that delay when we actually need the credentials. But I suspect it wasn't completely functioning as intended either way (at least in this case), so happy to just remove this logic.

PR #415, opened 2026-05-11
Release mutex per attachment, persist incrementally, allow stopSync to interrupt mid-batch

khawarizmus:
## Problem

Reported on Discord ([thread](https://discord.com/channels/1138230179878154300/1501962462910611597)).

1. `startSync`'s `asyncMap` wraps the entire per-attachment loop in `attachmentsService.withContext(...)`, so the attachment-service mutex is held for the whole batch.

2. `handleSync` iterates every active attachment serially with `await`, doing network I/O for each one inside that held mutex.

3. State changes are accumulated in a local list and committed once via a single `context.saveAttachments(updatedAttachments)` call at the end of the iteration.

4. Cancellation (`stopSync`) only re-evaluates the `takeWhile` predicate between batches so it cannot interrupt a running `asyncMap` step.

With ~20k queued downloads at ~200 ms each (the case in the report), it produces three observable failures:

- `stopSync()` can't complete until the in-flight `asyncMap` step finishes, i.e. until the full batch (~hours) drains.
- `saveFile()`, `deleteFile()`, `clearQueue()`, `expireCache()`, and `_processWatchedAttachments()` all call `withContext(...)` and are blocked behind the same mutex for the duration of the batch.
- Any consumer query against `attachments_queue` (e.g. a diagnostics page running `SELECT state, COUNT(*) ... GROUP BY state`) sees zero progress until the batch finishes and then a single atomic jump.

## Expected behavior

- `stopSync()` returns within one attachment's processing time, not one batch's.
- `saveFile()`, `deleteFile()`, and other queue mutations are not blocked behind an in-flight sync batch.
- `attachments_queue` reflects real-time progress, so consumer queries and `db.watch(...)` streams see incremental state changes instead of an end-of-batch commit.

## AI disclosure

This PR was created with the help of Opus 4.7 and Claude Code after reproducing the issue locally, and I’ve reviewed the resulting code changes.

### DIFF
--- a/packages/powersync/lib/src/attachments/sync/syncing_service.dart
+++ b/packages/powersync/lib/src/attachments/sync/syncing_service.dart
@@ -65,12 +65,21 @@ final class SyncingService {
         StreamGroup.merge<void>([attachmentChanges, manualTriggers])
             .takeWhile((_) => sub == _syncSubscription)
             .asyncMap((_) async {
-      await attachmentsService.withContext((context) async {
-        final attachments = await context.getActiveAttachments();
-        logger.info('Found ${attachments.length} active attachments');
-        await handleSync(attachments, context);
-        await deleteArchivedAttachments(context);
-      });
+      // Read the active set inside the mutex; release it before doing any
+      // per-attachment I/O so that concurrent `saveFile`/`deleteFile` and
+      // `_processWatchedAttachments` calls aren't blocked for the duration
+      // of the whole batch.
+      final attachments = await attachmentsService.withContext(
+        (context) => context.getActiveAttachments(),
+      );
+      logger.info('Found ${attachments.length} active attachments');
+
+      await handleSync(
+        attachments,
+        isActive: () => sub == _syncSubscription,
+      );
+
+      await attachmentsService.withContext(deleteArchivedAttachments);
     });
 
     _syncSubscription = sub = syncStream.listen(null);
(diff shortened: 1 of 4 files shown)


packages/powersync/lib/src/attachments/sync/syncing_service.dart
```diff
@@ -110,51 +119,72 @@ final class SyncingService {
...
+      if (isActive != null && !isActive()) {
+        logger.info('Sync cancelled; stopping iteration early');
+        return;
+      }
+
       logger.info(
         'Processing attachment ${attachment.id} with state: ${attachment.state}',
       );
+
+      Attachment? updated;
```
simolus3: Nit: Move this declaration into the `try` block, make it final and non-nullable (`final Attachment updated`). Then move the `saveAttachments` call to the end of the `try`, where the variable is guaranteed to have been assigned.

This lets us remove the `!= null` checks and the `continue` at the end of the try block.

packages/powersync/lib/src/attachments/sync/syncing_service.dart
```diff
@@ -110,51 +119,72 @@ final class SyncingService {
...
             break;
           case AttachmentState.queuedDelete:
             logger.info('Deleting [${attachment.filename}]');
-            updatedAttachments.add(await deleteAttachment(attachment, context));
+            // `deleteAttachment` needs a context (it removes the row in a
+            // transaction); briefly re-acquire the mutex for just this row.
+            updated = await attachmentsService.withContext(
+              (context) => deleteAttachment(attachment, context),
+            );
             break;
```
simolus3: Nit (since it's been like this before): We can remove those `break` statements here, there's no fall-through for non-empty cases in Dart.

packages/powersync/test/attachments/attachment_test.dart
```diff
@@ -325,6 +325,129 @@ void main() {
...
+          'INSERT INTO users (id, name, email, photo_id) VALUES (uuid(), ?, ?, ?)',
+          ['user-$i', 'user$i@example.com', 'photo-$i'],
+        );
+      }
+    });
+
+    // Sample partway through the batch: expect a partial count, neither 0
+    // (legacy behaviour: atomic end-of-batch commit) nor 20 (batch already
+    // finished, threshold too loose).
+    await Future<void>.delayed(const Duration(milliseconds: 700));
```
simolus3: Don't mention outdated behavior in comments. You can add a "// Regression test for https://github.com/powersync-ja/powersync.dart/pull/415" at the beginning of the test as additional context, but comments shouldn't describe what code no longer does :)

packages/powersync/test/attachments/attachment_test.dart
```diff
@@ -325,6 +325,129 @@ void main() {
...
+    final midBatchRow = await db.get(
+      'SELECT COUNT(*) AS c FROM attachments_queue WHERE state = ?',
+      [AttachmentState.synced.index],
+    );
```
simolus3: Can we refactor this test to use watched queries instead of relying lucky timing? E.g. if we had something like this:

```dart
  final syncedAttachmentsCounts = StreamQueue(db.watchUnthrottled('SELECT COUNT(*) ...', parameters: [AttachmentState.synced.index]).map((rs) => rs[0]['c'] as int));
```

Then we could do `await expectLater(syncedAttachmentsCounts, emitsThrough(allOf(greaterThan(0), lessThan(20))))` and hopefully get rid of the 750ms delay.

packages/powersync/test/attachments/attachment_test.dart
```diff
@@ -325,6 +325,129 @@ void main() {
...
+      for (var i = 0; i < 50; i++) {
+        await tx.execute(
+          'INSERT INTO users (id, name, email, photo_id) VALUES (uuid(), ?, ?, ?)',
+          ['user-$i', 'user$i@example.com', 'photo-$i'],
+        );
+      }
+    });
+
+    // Wait for the batch to be in flight.
+    await Future<void>.delayed(const Duration(milliseconds: 250));
```
simolus3: Instead of this fixed deleay, would it be possible to watch the local attachment state table and continue once we have an attachment being downloaded?

packages/powersync/test/attachments/attachment_test.dart
```diff
@@ -325,6 +325,129 @@ void main() {
...
+        await tx.execute(
+          'INSERT INTO users (id, name, email, photo_id) VALUES (uuid(), ?, ?, ?)',
+          ['savefile-user', 'savefile@example.com', attachment.id],
+        );
+      },
+    );
+    sw.stop();
+
+    // Before the fix this awaited the full batch (mutex held across all
+    // downloads). The fix releases the mutex between attachments.
```
simolus3: Describe the current state instead of old behavior (in this case I'm not sure we need a comment at all since the test's name describes what we want to assert).

simolus3 (commented):
I'm happy with these changes, thanks 👍 I was concerned about this potentially causing race conditions if we implicitly relied on the mutex being used everywhere, but since `SyncingService` is essentially an actor thanks to the `asyncMap` setup, this looks good to me.

My comments are mostly nits or related to the tests: IMO, we should avoid relying on magic `Future.delayed` calls as much as possible since that makes tests less reliable (I see that one test became flaky in the initial CI run for this PR for instance).

packages/powersync/lib/src/attachments/sync/syncing_service.dart
khawarizmus: Fixed

packages/powersync/lib/src/attachments/sync/syncing_service.dart
khawarizmus: Fixed

packages/powersync/test/attachments/attachment_test.dart
khawarizmus: Addressed

packages/powersync/test/attachments/attachment_test.dart
khawarizmus: Fixed

packages/powersync/test/attachments/attachment_test.dart
khawarizmus: We used the same suggestion as above. using `db.watchUnthrottled` and then `expectLater`

packages/powersync/test/attachments/attachment_test.dart
khawarizmus: Addressed

khawarizmus:
Thank you for the review @simolus3 

I have updated the PR. Please let me know what you think

simolus3 (approved):
LGTM!

PR #416, opened 2026-05-12
Improve actions setup, add static analysis

simolus3:
This adopts [zizmor](https://zizmor.sh/) as a static analysis tool for GitHub actions workflows in this repository, and fixes most lints. I'm mainly opening this to start a discussion - is this a tool we want to use or are the reports too noisy to be useful for us?

In this repository, the default configuration flagged:

1. Using actions without pinned refs: Zizmor warns about using actions like `actions/checkout@v6` since `v6` is a mutable tag that could change between runs. GitHub allows us to forbid mutable references altogether, too. IMHO, this is a bit too strict by default:
    - Is someone managed to attack `actions/checkout`, we'd have bigger problems. So I think explicitly using the latest version makes sense.
    - It's a similar story for Homebrew and Dart. We use whatever they ship anyway, so pinning an action doesn't help us at all.
    - `subosito/flutter-action` is community-maintained and adding an explicit ref can make sense for those.
2. To keep up with changes to ref-pinned actions dependencies, this configures dependabot. Zizmor also checks these configurations, and enforces a cooldown period for instance.
3. Workflows without an explicit `permissions` block since the default value is too broad. This makes sense to fix, and we also get warnings from GitHub about this.
4. `actions/checkout` uses without `persist-credentials: false`. That's a good point IMO, we don't need these credentials.
5. `${{ ... }}` expansions in scripts since it could "lead to running attacker-controlled code". This is a false-positive in `publish.yml` which was flagged here, but it can still be cleaned up with relatively little effort.

Zizmor can also be configured with a more strict ruleset (`zizmor --persona pedantic .`). In this repository, that additionally flags:

- Jobs without a `concurrency` setting (only affects the publishing workflow in this repository which is harmless).
- Jobs without an explicit name (this doesn't feel that useful to me).
- Permissions without a YAML comment next to them explaining why they're necessary (this one seems helpful).

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -5,7 +5,7 @@ runs:
   using: composite
   steps:
     - name: Install Flutter
-      uses: subosito/flutter-action@v2
+      uses: subosito/flutter-action@1a449444c387b1966244ae4d4f8c696479add0b2 # v2.23.0
       with:
         flutter-version: "3.x"
         channel: "stable"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 7
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,7 +14,7 @@ permissions: {}
 jobs:
   zizmor:
     permissions:
-      security-events: write
+      security-events: write # Needed to upload findings as code scanning results.
     runs-on: ubuntu-latest
     steps:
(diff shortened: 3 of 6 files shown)


rkistner (approved):
I'm happy with these checks and changes.

Just to check - it looks like this also adds dependabot config to semi-automatically update github action versions?

simolus3:
> it looks like this also adds dependabot config to semi-automatically update github action versions?

Yes, I should have mentioned that in the PR description. My thinking is that dependabot would create PRs to update refs since we no longer get updates for new minor releases (unlike with `@v2` which resolves to the latest compatible version on each run). That still has some risk I suppose since these PRs would run with the context of the repository, but the empty permission set should cover that (and we get to check changes to actions before merging).

stevensJourney (approved):
I like these checks, I'd be happy if we added these to our other repos :)

PR #421, opened 2026-05-26
Refactor shutdown procedure of sync isolates

### LINKED ISSUE #419: Bug: sync isolate shutdown doesn't await database.close() — leaks connection pool isolates on every disconnect cycle
## Summary

In `powersync` v2.1.0, the sync isolate's `shutdown()` function calls `database.close()` without `await`, then immediately calls `Isolate.current.kill()`. This kills the isolate before the connection pool has finished closing its worker isolates, leaking them on every connect/disconnect cycle.

As my application might open and close the database in some cases, the cycles could be accumulated very quickly.

After enough cycles, the leaked isolates accumulate until the app hits the iOS memory limit and is killed by the OS.

## Root Cause

**File:** `lib/src/database/native/native_powersync_database.dart`, inside `_syncIsolate` → `shutdown()`

```dart
Future<void> shutdown() {
  if (!shutdownCompleter.isCompleted) {
    shutdownCompleter.complete(Future(() async {
      await openedStreamingSync?.abort();
      crudUpdateController.close();
      database.close();          // ← NOT AWAITED
      rPort.close();
      Isolate.current.kill();    // ← kills isolate before close() finishes
    }));
  }
  return shutdownCompleter.future;
}
```

SqliteDatabase.close() needs to shut down its connection pool workers (up to 5 reader isolates). Without await, Isolate.current.kill() fires before the pool finishes cleanup, orphaning those worker isolates.

One-Line Fix
```dart
- database.close();
+ await database.close();
```

Environment
powersync: 2.1.0
sqlite_async: 0.14.1
Platform: iOS (iPhone, ~1.85 GB memory limit)
Database size: ~1 GB
Usage pattern: repeated sign-in/sign-out or business-switching cycles

Reproduction
Open a PowerSync database (~1 GB)
Connect to PowerSync
Close/disconnect the database
Repeat steps 1-3 multiple times
Observe isolate count growing in Dart DevTools — worker isolates from previous cycles are never cleaned up
After ~3-5 cycles on iOS, the app is killed by the OS (memory limit exceeded)

Additional Context
The TODO comment already in the code hints at this:

// TODO: If we closed our resources properly, this wouldn't be necessary...
Isolate.current.kill();
The main isolate's waitForShutdown() correctly awaits receivedIsolateExit.future, but by the time the sync isolate exits, its database.close() hasn't completed — the connection pool workers survive as orphans.

I patched this locally (single await added) and confirmed it resolves both the slow close times and the memory leak. Would appreciate an official fix in a patch release given the severity on iOS devices with large databases.

simolus3:
Starting from version 2.0.0 of the SDK, the sync isolate opens a proper second reference to the database pool used by the main app. This includes spawning additional isolates to dispatch queries, and those isolates aren't properly closed because the sync isolate doesn't await `database.close()` before killing itself.

The fix for that alone would be simple, add an `await` and things should work. However, this seemed like a good opportunity to investigate why `Isolate.current.kill()` is even necessary. It's kind of a code smell since we're supposed to have cleared all resources keeping the isolate alive at that point. This PR fixes several issues and refactors how the sync isolate is managed.

1. Instead of sending untyped lists with string commands between the sync isolate and the main app, this introduces `SyncIsolateToClientMessageType` and `ClientToSyncIsolateMessageType` enums documenting available messages.
2. Direct `SendPort.send` invocations are replaced with an extension type exposing messages as well-typed methods.
3. We no longer send crud updates from the main isolate to the sync isolate (the `sqlite_async` package takes care of that for us since the 2.0.0 update, we can just listen on `onChange`).
4. Avoid the use of `Future.delayed` in the streaming sync implementation: It internally starts a `Timer`, which will keep the isolate alive until the timer expires. There's no way to "unregister" a listener on futures, so this can't be cancelled. This refactors `_delayRetry` to properly clean up the timer when the sync is aborted. This makes sync isolate exit much quicker.
5. Add `IsolateResultCollection` as a wrapper around `IsolateResult`s to be able to abort isolate results properly (this didn't turn out to be the issue, but is still an improvement: We shouldn't have to wait for outstanding crud uploads before shutting down the sync isolate).

AI use disclaimer: `isolate_completer_test.dart` was generated by Claude, but manually revised.

Closes https://github.com/powersync-ja/powersync.dart/issues/419.


stevensJourney (approved):
This looks like a great update.

PR #422, opened 2026-05-27
Update `sqlite_async`, prepare release

simolus3:
This updates the `sqlite_async` package to its latest 0.8.0 version:

- Update the `sqlite3.wasm` build (to update the underlying SQLite version).
- Use strings instead of `Uri` instances.
- Update to the latest version of `sqlite_async`, and support the latest versions of `hooks` and `code_assets`
- Prepare a minor release.

## Product visibility

A noteworthy new feature we get inherit from the `sqlite3_web` and `sqlite_async` update is that OPFS is now supported on Safari without additional headers (documentation update: https://github.com/powersync-ja/powersync-docs/pull/474).

### DIFF
--- a/packages/powersync/CHANGELOG.md
+++ b/packages/powersync/CHANGELOG.md
@@ -1,6 +1,9 @@
-## 2.1.1 (unreleased)
+## 2.1.1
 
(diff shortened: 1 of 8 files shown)


stevensJourney (approved):
🚀

PR #424, opened 2026-06-02
Support custom HTTP clients

simolus3:
There are a number of cases where users might want to use custom HTTP clients with the PowerSync SDK, such as:

- Adding custom request headers for a self-hosted PowerSync service behind an authenticating proxy (https://github.com/powersync-ja/powersync.dart/pull/420).
- Trusting additional self-signed certificates (https://github.com/powersync-ja/powersync.dart/pull/420#issuecomment-4575927758).
- Transforming `/sync/stream` responses for the middleware-based encryption approach.
- Using more efficient HTTP client implementations (like [cronet_http](https://pub.dev/packages/cronet_http) or [cupertino_http](https://pub.dev/packages/cupertino_http)).

This PR adds support for this through an `httpClient` field on `SyncOptions`. It can be set to a `Client Function()` acting as a factory for HTTP clients. On native platforms, we send that factory function through a send port to reconstruct a client on the other end.

On the web, the whole thing is a little tricky. We can't send Dart objects to a worker, so this serializes HTTP requests and responses via `postMessage` calls:

1. When the sync worker wants to issue an HTTP request, it sends the method, url, headers and body to a client.
2. The client can then use the custom HTTP client to run the request, and replies with the status code and headers.
3. The sync worker then requests chunks of the response, essentially mirroring the `AsyncIterator` interface through message ports.

We only do this when custom clients are installed, since `postMessage` might add overhead (probably not too much since we can transfer blobs though).

Note: According to mdn, `ReadableStream`s are supposed to be transferrable, which would have made the whole implementation much simpler. But also according to mdn, `ReadableStream`s aren't transferrable on Safari :/

TODOs:

- [x] Add unit test for native platforms.
- [x] Manually test this on the web.
- [x] On the web, recover from the client tab being closed while it's serving an HTTP request.

### DIFF
--- a/packages/powersync/CHANGELOG.md
+++ b/packages/powersync/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.3.0 (unreleased)
+
+- Add `SyncOptions.httpClient` to `SyncOptions`. It can be set to make PowerSync use a custom HTTP client when connecting to the PowerSync Service.
+
 ## 2.2.0
 
 - Fix `disconnect()` call leaving stale database isolates behind.
--- a/packages/powersync/lib/src/database/native/native_powersync_database.dart
+++ b/packages/powersync/lib/src/database/native/native_powersync_database.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:isolate';
+import 'package:http/http.dart';
 import 'package:meta/meta.dart';
 
 import 'package:logging/logging.dart';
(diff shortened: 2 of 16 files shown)


T0b1i:
I tested this PR against our self-hosted setup and it solves our CA-trust problem. Returning a package:http IOClient built on an HttpClient with a custom SecurityContext carrying our CA, via the Client Function() factory, connects fine and is fast. The factory covers the self-signed-cert case for us. I also tried a cronet_http client but it was extremely slow in the spawned sync isolate, and I haven't figured out why. We're going with dart:io, which works well.

Is there a concrete plan for when this feature will be officially released?

Thanks for the great work on this!

Chriztiaan (approved):

PR #425, opened 2026-06-03
Improve tab close detection for shared sync worker

simolus3:
Currently, sync workers detect closed tabs by listening on `database.closed` for a database backed by `sqlite3_web` (which internally uses navigator locks to detect closed tabs).

This kind of works, but can lead to deadlocks when a tab is closed while it's serving a request (like a `fetchCredentials` or `requestDatabase` call) that would never complete.

To improve reliability, this adopts the usual navigator locks pattern to our `WorkerCommunicationChannel` as well:

1. Before requesting a sync iteration, tabs acquire a randomly-named lock, and they include that lock name in the request.
2. Like before, the shared sync worker will mark a tab as closed once this lock is available to it. But starting from this PR, it also throws for all pending requests (they would never complete before).

I've also added a small test for the shared sync worker in `sync_worker_test.dart`. It emulates a shared worker by manually passing ports from a `MessageChannel` to it. I've also tested this in a demo app (and actually found an issue in `sqlite3_web`, I'll fix that too).

### DIFF
--- a/packages/powersync/lib/src/web/sync_controller.dart
+++ b/packages/powersync/lib/src/web/sync_controller.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:js_interop';
 
+import 'package:meta/meta.dart';
 import 'package:powersync/src/web/worker_utils.dart';
 import 'package:sqlite_async/web.dart';
 import 'package:web/web.dart';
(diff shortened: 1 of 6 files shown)


Chriztiaan:
I found the following issue which I am not sure if it's a regression:

Launch `supabase-todolist` demo for `web` in two chrome tabs. Both show connected state, changes to the source database sync down to both tabs. I refresh the first tab, it never reaches a connected state and the second tab still shows connected. Changes to the source database don't sync down to either tab. If I close the first tab and refresh the second tab the data does then sync down.

simolus3:
> I found the following issue which I am not sure if it's a regression:

Thanks for trying this :) This was a `sqlite3_web` issue I've hinted at in the PR description (I should have been clearer about that, sorry if that wasted time). In-flight requests to the database would block forever when the database is closed instead of erroring. Upgrading to 0.8.1 and rebuilding the workers makes them recover from this.

Chriztiaan (approved):

PR #432, opened 2026-06-13
fix(powersync): use read lock for BucketStorage.select()

rumitvn:
Closes #384

### Problem
`BucketStorage.select()` is documented "Use only for read statements", but it calls `_internalDb.execute()`, which acquires a **write lock**. Every caller (`getClientId`, `updateLocalTarget`, `hasCrud`, `getCrudBatch`, …) runs read-only `SELECT`s, so taking a write lock needlessly serializes these reads behind the write lock — notably on web / `SqliteConnectionPool`, where reads and writes can otherwise run concurrently.

@simolus3 confirmed in the issue that this is a bug.

### Fix
Route `select()` through `getAll()`, which acquires a **read lock**, matching the method's documented read-only contract. All existing callers are plain `SELECT`s, so behaviour is unchanged apart from the correct lock type.

### Notes
- One-line behavioural change + CHANGELOG entry (`## 2.3.1 (unreleased)`).
- No public API change.

### DIFF
--- a/packages/powersync/CHANGELOG.md
+++ b/packages/powersync/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.3.1 (unreleased)
+
+- Fix `BucketStorage.select()` acquiring a write lock for read-only queries by using `getAll()` instead of `execute()`.
+
 ## 2.3.0
 
 - Add `SyncOptions.httpClient` to `SyncOptions`. It can be set to make PowerSync use a custom HTTP client when connecting to the PowerSync Service.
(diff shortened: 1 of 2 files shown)


packages/powersync/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.3.1 (unreleased)
+
+- Fix `BucketStorage.select()` acquiring a write lock for read-only queries by using `getAll()` instead of `execute()`.
```
simolus3: No need to add implementation details to public changelogs

```suggestion
- Fix `BucketStorage.select()` acquiring a write lock for read-only queries.
```

simolus3 (dismissed):
Thanks! These changes look good to me, could you also rebase to resolve the conflict in the changelog?

simolus3 (approved):
LGTM, we'll need another approval now.

LucDeCaf (approved):

PR #441, opened 2026-07-15
Don't create sync status entries in Dart

### LINKED ISSUE #440: hasSynced incorrectly resurrects to true after disconnectAndClear() + connect(), causing waitForFirstSync() to resolve on an empty database
## Environment

- `powersync: 2.3.1` (Flutter, native/Android, Rust sync client)
- powersync-service 1.23.3, sync streams (edition 3, all `auto_subscribe: true`)

## Summary

After `disconnectAndClear()` (logout) followed by `connect()` (re-login in the same app session), `SyncStatus.hasSynced` flips back to `true` **before any data has been downloaded**. `waitForFirstSync()` therefore resolves immediately while all synced tables are still empty. Apps that gate their post-login navigation on first sync land on empty screens; data appears seconds later when the real first checkpoint completes.

## Root cause

Two pieces interact:

**1. `manuallyChangeSyncStatus` can never clear `lastSyncedAt`** (`lib/src/sync/connection_manager.dart`):

```dart
lastSyncedAt: status.lastSyncedAt ?? currentStatus.lastSyncedAt,
hasSynced: status.lastSyncedAt != null
    ? true
    : status.hasSynced ?? currentStatus.hasSynced,
```

`disconnectAndClear()` ends with `setStatus(SyncStatus(lastSyncedAt: null, hasSynced: false))`, but because of the `??` fallback, `lastSyncedAt: null` cannot overwrite the previous value. Result after clear: `hasSynced=false` but the stale `lastSyncedAt` is retained in `currentStatus` (observable in the status stream).

**2. The sync isolate shutdown handler re-emits the stale value** (native `connectInternal`, `waitForShutdown` in `lib/src/database/native/native_powersync_database.dart`):

```dart
// Clear status apart from lastSyncedAt
setStatus(SyncStatus(lastSyncedAt: currentStatus.lastSyncedAt));
```

On the next `connect()`, the previous iteration's shutdown emits this status with the stale non-null `lastSyncedAt`, which the merge logic above turns into `hasSynced: true` — before the new sync iteration has downloaded anything.

## Logs (annotated)

Logout — note `hasSynced=false` but `lastSyncedAt` retained, while the DB-level value is correctly cleared:

```
disconnectAndClear start
status: connected=false hasSynced=true  lastSyncedAt=2026-07-15 13:50:58.000
status: connected=false hasSynced=false lastSyncedAt=2026-07-15 13:50:58.000   <- stale lastSyncedAt survives
disconnectAndClear done in 60ms hasSynced=false db_last_synced_at=null         <- SELECT powersync_last_synced_at() is null
```

Re-login / `connect()` — `hasSynced` resurrects before download starts, stays `true` through the whole first download:

```
[PowerSync] FINE: Sync Isolate exit
status: connected=false connecting=false hasSynced=true lastSyncedAt=2026-07-15 13:50:58.000  <- resurrected, 0 bytes downloaded
[PowerSync] INFO: Starting Rust sync iteration
status: connected=true downloading=true hasSynced=true lastSyncedAt=13:50:58 progress=0 / 4333
...
status: connected=true downloading=true hasSynced=true lastSyncedAt=13:50:58 progress=4333 / 4333
status: connected=true downloading=false hasSynced=true lastSyncedAt=2026-07-15 13:52:22.000  <- real first checkpoint
```

## Expected

After `disconnectAndClear()`, `hasSynced` should remain `false` (and `lastSyncedAt` `null`) until a new full sync checkpoint has been applied, so `waitForFirstSync()` blocks until local data actually exists — matching the DB-level `powersync_last_synced_at()` semantics.

## Actual

`waitForFirstSync()` resolves immediately on reconnect while all synced tables are empty.

## Repro

1. Connect and complete a first sync.
2. `await db.disconnectAndClear();`
3. `await db.connect(connector: ...);`
4. `await db.waitForFirstSync();` → resolves instantly; querying any synced table returns 0 rows.

## Workaround

Gate on the database truth instead of the in-memory status: treat first sync as complete only when `SELECT powersync_last_synced_at()` is non-null (it is correctly cleared by `powersync_clear`), and re-check it on status-stream events instead of using `waitForFirstSync()`.

simolus3:
With the Rust sync client, `SyncStatus` is supposed to to be managed by the core extension (apart from upload and download errors).

Still, we had a few places where we'd emit a fake sync status resetting all fields apart from `lastSyncedAt`. This is wrong, it resets sync streams even though those are still there after disconnecting and in the case of `disconnectAndClear`, it could actually cause us to report a database as synced when it's not.

To fix this, this PR:

1. Makes it much more annoying to create `SyncStatus` entries by marking all fields as required. The only place in the SDK (apart from tests) where we create a sync status now is when mapping from core state to Dart, which is the only place where we're supposed to do that.
2. Explicitly loads the offline sync status after disconnecting or clearing the database to make sure the reported sync status reflects the actual state.

AI use: Changes are manual, I ran a `/review` in Claude Code and manually fixed findings before marking this PR as ready for review.

Closes https://github.com/powersync-ja/powersync.dart/issues/440.

### DIFF
--- a/packages/powersync/lib/src/database/native/native_powersync_database.dart
+++ b/packages/powersync/lib/src/database/native/native_powersync_database.dart
@@ -74,8 +74,6 @@ final class NativePowerSyncDatabase extends BasePowerSyncDatabase {
       receiveExit.close();
       mutexServer.handleChildIsolateExit();
 
-      // Clear status apart from lastSyncedAt
-      setStatus(SyncStatus(lastSyncedAt: currentStatus.lastSyncedAt));
       abort.completeAbort();
     }
 
--- a/packages/powersync/lib/src/sync/mutable_sync_status.dart
+++ b/packages/powersync/lib/src/sync/mutable_sync_status.dart
@@ -49,7 +49,7 @@ final class MutableSyncStatus {
     streams = status.streams;
   }
 
-  SyncStatus immutableSnapshot({bool setLastSynced = false}) {
+  SyncStatus immutableSnapshot() {
     return SyncStatus(
       connected: connected,
       connecting: connecting,
@@ -58,7 +58,6 @@ final class MutableSyncStatus {
       downloadProgress: downloadProgress?.asSyncDownloadProgress,
       priorityStatusEntries: UnmodifiableListView(priorityStatusEntries),
       lastSyncedAt: lastSyncedAt,
-      hasSynced: setLastSynced ? lastSyncedAt != null : null,
       uploadError: uploadError,
       downloadError: downloadError,
       streamSubscriptions: streams,
@@ -68,7 +67,7 @@ final class MutableSyncStatus {
 
 final class SyncStatusStateStream {
   final MutableSyncStatus status = MutableSyncStatus();
-  SyncStatus _lastPublishedStatus = const SyncStatus();
+  SyncStatus _lastPublishedStatus = const SyncStatus.uninitialized();
 
   final StreamController<SyncStatus> _statusStreamController =
       StreamController<SyncStatus>.broadcast();
--- a/demos/supabase-todolist-drift/lib/powersync/powersync.dart
+++ b/demos/supabase-todolist-drift/lib/powersync/powersync.dart
@@ -54,8 +54,9 @@ final _syncStatusInternal = StreamProvider<SyncStatus?>((ref) {
 });
 
 final syncStatus = Provider((ref) {
-  // ignore: invalid_use_of_internal_member
-  return ref.watch(_syncStatusInternal).value ?? const SyncStatus();
+  return ref.watch(_syncStatusInternal).value ??
+      // ignore: invalid_use_of_internal_member
+      const SyncStatus.uninitialized();
 });
 
 @riverpod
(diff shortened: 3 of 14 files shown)


packages/powersync/lib/src/database/native/native_powersync_database.dart
```diff
@@ -74,8 +74,6 @@ final class NativePowerSyncDatabase extends BasePowerSyncDatabase {
       receiveExit.close();
       mutexServer.handleChildIsolateExit();
 
-      // Clear status apart from lastSyncedAt
-      setStatus(SyncStatus(lastSyncedAt: currentStatus.lastSyncedAt));
```
simolus3, line 78: Note that the core extension emits an `UpdateSyncStatus` instruction when stopping a sync iteration, which made this useless. As mentioned in the PR description, it also clears fields which aren't supposed to be cleared.

packages/powersync/lib/src/sync/mutable_sync_status.dart
```diff
@@ -49,7 +49,7 @@ final class MutableSyncStatus {
     streams = status.streams;
   }
 
-  SyncStatus immutableSnapshot({bool setLastSynced = false}) {
+  SyncStatus immutableSnapshot() {
```
simolus3, line 52: The `hasSynced` field is unconditionally backed by `lastSyncedAt != null` now.

The only way for `hasSynced` to be `null` is when we're still opening the database and don't know what the offline sync status is. I've added a separate constructor for that state, simplifying this in the rest of the SDK.

Chriztiaan (approved):

PR #446, opened 2026-07-27
Reduce log severity from attachment queue service

### LINKED ISSUE #408: Attachment queue logs "is already synced" for every row on every sync tick
Context: https://discord.com/channels/1138230179878154300/1495760178803507271

## Summary

The attachment queue emits an info-level log line for every non-archived attachment on every periodic sync tick, including rows already in the synced state. With more than a handful of attachments, this becomes constant log spam that is hard to filter without silencing the whole PowerSync logger.

The JS SDK handles the same path silently. Aligning Dart with JS would fix the spam with no user-facing API change.

## Where this happens

The periodic pipeline loads every non-archived attachment and hands them all to `handleSync` which then logs a line for every attachment, including synced and archived branches:

[packages/powersync/lib/src/attachments/sync/syncing_service.dart#L118-L153](https://github.com/powersync-ja/powersync.dart/blob/main/packages/powersync/lib/src/attachments/sync/syncing_service.dart#L118-L153)

The periodic trigger itself also logs at info every tick:

[syncing_service.dart#L79-L84](https://github.com/powersync-ja/powersync.dart/blob/main/packages/powersync/lib/src/attachments/sync/syncing_service.dart#L79-L84)

```dart
_periodicSubscription = Stream<void>.periodic(period, (_) {}).listen((_) {
  logger.info('Periodically syncing attachments');
  triggerSync();
});
```

## How the JS SDK handles the same path

`processAttachments` only emits logs for actionable states; synced/archived rows fall through default: break silently:

[packages/common/src/attachments/SyncingService.ts#L44-L67](https://github.com/powersync-ja/powersync-js/blob/main/packages/common/src/attachments/SyncingService.ts#L44-L67)

The only info logs in JS are inside `uploadAttachment` / `downloadAttachment`; they fire only when actual work happens. No per-tick noise.

## Proposed fix

Bring the Dart `handleSync` in line with the JS `processAttachments`:

- Drop (or demote to `Level.FINE`) the `logger.info('Processing attachment …')` header inside the loop.

- Drop the `synced` and `archived` `logger.info(...)` lines. Synced/archived rows should pass through silently, like JS.

- Consider demoting the per-tick diagnostic lines (`Periodically syncing attachments`, `Starting handleSync with N attachments`, `Found N active attachments`) from `info` to `fine/finer`.

simolus3:
Currently, the syncing logic of the attachment system logs the same stuff multiple times and with a severity that is annoyingly high. This PR reduces visible logs:

- Reduce periodic log messages from the normal attachment scanning interval to level `fine` and lower, meaning that they wouldn't get logged by default loggers.
- Since `downloadAttachment` and friends log a message before they do anything, we don't need to log almost the same message before calling them.
- When an error handler determines that an attachment should be archived instead of retried, we sometimes didn't log the original error. Fix that by logging the error first.

In other words, we only log when the attachment service actually does stuff (downloading, uploading, deleting). Its internal state can be observed with a logger set to level `fine` and lower for debugging purposes, but this wouldn't happen by default.

Closes #408.

### DIFF
--- a/packages/powersync/CHANGELOG.md
+++ b/packages/powersync/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.3.4 (unreleased)
+
+- Reduce log noise from attachments service ([#408](https://github.com/powersync-ja/powersync.dart/issues/408)).
+
 ## 2.3.3
 
 - Restore the ability to call `disconnect()` on closed databases.
--- a/packages/powersync/lib/src/attachments/attachment_queue_service.dart
+++ b/packages/powersync/lib/src/attachments/attachment_queue_service.dart
@@ -180,6 +180,7 @@ base class AttachmentQueue {
       var previouslyConnected = _db.currentStatus.connected;
       _syncStatusSubscription = _db.statusStream.listen((status) {
         if (!previouslyConnected && status.connected) {
+          _logger.fine('Scanning attachments after connecting');
           _syncingService.triggerSync();
         }
 
--- a/packages/powersync/lib/src/attachments/sync/syncing_service.dart
+++ b/packages/powersync/lib/src/attachments/sync/syncing_service.dart
@@ -205,11 +205,12 @@ final class SyncingService {
         e,
         st,
       );
-      if (errorHandler != null) {
-        final shouldRetry =
-            await errorHandler!.onUploadError(attachment, e, st);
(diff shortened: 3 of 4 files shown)


khawarizmus (dismissed):
LGTM

Chriztiaan (approved):

PR #465, opened 2026-08-24
Sync client: Improve handling abort requests

simolus3:
To prepare the sync client implementation for the new checkpoint requests protocol, this cleans up some async code:

1. Consistently pass an abort controller to methods that need it. This makes impossible states unrepresentable: We're guaranteed to have an abort controller when the sync client runs, but we have a few  `!= null` checks on the field. Passing a non-nullable abort controller cleans that up.
2. Allow requests to write checkpoints to be aborted: Calling `disconnect()` no longer waits for a write checkpoint if an upload task is currently running. Arguably not that important today, but more relevant for checkpoint requests which are posted more frequently.
3. Use scoped futures in an abort controller: A downside of representing abort requests as a single `Future<void>` is that it's impossible to remove listeners from a future in Dart. For a long-running sync iteration that is never aborted, every download iteration and retry delay adds a listener to the future that is never removed. While this leak isn't terrible, re-posting checkpoint requests until they're applied will only make things worse. This adds the `AbortController.scoped` method, which passes a fresh abort future for a smaller unit of async work. We can forget about that future once the async unit completes, fixing the memory leak.
4. In the sync client, don't store `_activeCrudUpload` as a field. Instead, follow what we do on most other SDKs (I checked JS, Kotlin and Swift) and await the whole upload loop. This makes it easier to add new concurrent sync tasks in the future (like the retry loop for checkpoint requests).
5. Don't inform the core extension about completed uploads when no upload has actually been completed, matching [what we do on JS](https://github.com/powersync-ja/powersync-js/blob/a875d48b8e99e0ce9999d3f8b305e56d2f554475/packages/shared-internals/src/client/sync/stream/AbstractStreamingSyncImplementation.ts#L280-L288).

AI use: Reviewed with Claude Code.

### DIFF
--- a/packages/powersync/lib/src/sync/streaming_sync.dart
+++ b/packages/powersync/lib/src/sync/streaming_sync.dart
@@ -311,7 +311,7 @@ class StreamingSyncImplementation implements StreamingSync {
     return response;
   }
 
-  Future<http.StreamedResponse?> _postStreamRequest(
+  Future<http.StreamedResponse> _postStreamRequest(
     Object? data, {
     required Future<void> onAbort,
   }) async {
@@ -456,18 +456,14 @@ final class _ActiveRustStreamingIteration {
     return streamFromFutureAwaitInCancellation(
       sync._postStreamRequest(data, onAbort: onAbort),
     ).asyncExpand<SyncEvent>((response) async* {
-      if (response == null) {
-        return;
-      } else {
-        yield ConnectionEvent.established;
+      yield ConnectionEvent.established;
 
-        final contentType = response.headers['content-type'];
-        final isBson = contentType == 'application/vnd.powersync.bson-stream';
+      final contentType = response.headers['content-type'];
+      final isBson = contentType == 'application/vnd.powersync.bson-stream';
 
-        yield* (isBson ? response.stream.bsonDocuments : response.stream.lines)
-            .map(ReceivedLine.new);
-        yield ConnectionEvent.end;
-      }
+      yield* (isBson ? response.stream.bsonDocuments : response.stream.lines)
+          .map(ReceivedLine.new);
+      yield ConnectionEvent.end;
     });
   }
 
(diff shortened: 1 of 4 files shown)


packages/powersync/lib/src/sync/streaming_sync.dart
```diff
@@ -142,57 +129,58 @@ class StreamingSyncImplementation implements StreamingSync {
...
+      // On error, wait a little before retrying
+      // When aborting, don't wait
+      if (!abort.aborted && delayNextIteration) {
+        await _delayRetry(abort);
+      }
     }
   }
 
-  Future<void> _crudLoop() async {
-    await _uploadAllCrud();
```
simolus3, line 194: I think this initial upload isn't necessary: When we start a sync iteration, we observe `ps_crud` changes and deliberately add an empty event ([native](https://github.com/powersync-ja/powersync.dart/blob/75cbf540fd84b88dc1b42168d5922a4e8a87f9da/packages/powersync/lib/src/database/native/native_powersync_database.dart#L299-L300) using `onChange` with the default `triggerImmediately: true`, [web](https://github.com/powersync-ja/powersync.dart/blob/75cbf540fd84b88dc1b42168d5922a4e8a87f9da/packages/powersync/lib/src/web/sync_worker.dart#L310)).

packages/powersync/lib/src/sync/streaming_sync.dart
```diff
@@ -290,13 +278,14 @@ class StreamingSyncImplementation implements StreamingSync {
       'write-checkpoint2.json?client_id=$clientId',
     );
 
-    Map<String, String> headers = {
-      'Content-Type': 'application/json',
```
simolus3, line 294: `Content-Type` is supposed to describe the body of the request itself, but there is none since this is a `GET`. Changed this to `Accept`.

packages/powersync/lib/src/sync/streaming_sync.dart
```diff
@@ -335,22 +324,14 @@ class StreamingSyncImplementation implements StreamingSync {
...
-    request.headers['Authorization'] = "Token ${credentials.token}";
     request.headers['Accept'] = '$bson;q=0.9,$ndJson;q=0.8';
-    request.headers.addAll(_userAgentHeaders);
 
     request.body = convert.jsonEncode(data);
 
     final res = await _client.send(request);
-    if (aborted) {
-      return null;
-    }
```
simolus3, line 353: `_client.send()` throws when we abort the request before receiving a response, so catching that exception and reocgnizing it (in `isAbortException`) removes another `!= null` check.

rkistner (approved):

PR #468, opened 2026-08-25
Explicit checkpoint requests

simolus3:
This adds support for explicit checkpoint requests through `PowerSyncDatabase.requestCheckpoint()`. Checkpoint requests can be issued when the database is connected, and used to wait until the serverside snapshot at the time of the checkpoint has finished syncing.

The implementation here mostly follows the existing JavaScript and Kotlin implementation. A Dart-specific concern is that requesting checkpoints from workers and the isolate requires an explicit protocol.

https://github.com/user-attachments/assets/3b8ec46d-604f-4f84-bac9-e7c87e7785f2

AI used: I used Claude Code to help add this to the demo app and to review this.


packages/powersync/lib/src/sync/connection_manager.dart
```diff
@@ -278,6 +290,16 @@ final class ConnectionManager {
     return _SyncStreamImplementation(this, name, parameters);
   }
 
+  Future<CheckpointRequest> requestCheckpoint() async {
+    return _activeGroup.syncConnectMutex.lock(() async {
+      final client = _abortActiveSync?.$2;
+      checkConnectedWithRequestsMode();
```
LucDeCaf, line 296: This looks a little bit racey, but I'm assuming since Dart is single-threaded it's probably fine. Just wanted to mention it.

packages/powersync/lib/src/sync/streaming_sync.dart
```diff
@@ -27,7 +27,13 @@ import 'sync_status.dart';
 
 typedef SubscribedStream = ({String name, String parameters});
 
-abstract interface class StreamingSync {
+/// A superinterface for [StreamingSync] providing methods that can be called
+/// in remote contexts.
+abstract interface class RemoteStreamingSyncHandle {
```
LucDeCaf, line 34: Lacking context here, but why do we need a new interface for this instead of reusing `StreamingSync`? Does it have to do with inter-isolate calls?

LucDeCaf (commented):
This seems correct logic-wise, but I have some comments/questions since I'm not really that familiar with Dart.

packages/powersync/lib/src/sync/streaming_sync.dart
simolus3: That, and it's also the fact that the `StreamingSync` interface isn't really used:

- The `statusStream` getter is used directly on the native/web database implementation to forward events.
- Similarly, `streamingSync()` and `abort()` are meant to be called directly.
- `updateSubscriptions()` is also called directly on the database implementation (it receives a stream of updates to listen to).

IMO, having both implementations return a `StreamingSync` and then managing the rest in shared code would be cleaner. I'll add a todo comment for that.

packages/powersync/lib/src/sync/connection_manager.dart
simolus3: `_abortActiveSync` must only be interacted with in a `syncConnectMutex.lock` scope to avoid race conditions, there is one here as well. This guarantees the client is not changed or aborted while we're in `requestCheckpoint()`. I'm happy to add a clarifying comment if that helps, which potential data race did you see?

LucDeCaf: Oh wait, I somehow missed the fact that the call was wrapped in a lock context. Nevermind 🫠

LucDeCaf (approved):

PR #473, opened 2026-09-09
Bundle the PowerSync agent skill

### LINKED ISSUE #472: Include agent skills for users?
PowerSync agent skills are distributed as npm package, but Dart developers might not necessarily have npm installed.

Dart provides the [skills package](https://pub.dev/packages/skills), which we can use to bundle skills with the `powersync` package. Once users depend on that package, they can run `dart run skills@ get` and would be prompted to install PowerSync-related skills.

Distributing skills that way makes them easier to discover for Dart users, but it would mean that we'd need to sync them from the [main repository](https://github.com/powersync-ja/agent-skills) to here. That's fairly easy to automate though.

cc @bean1352

bean1352:
This bundles `v1.5.0` of the `powersync` skill from `powersync-ja/agent-skills` into `packages/powersync/skills/`, so users of this package can install it with `dart run skills@ get`.

Only the `skills/powersync` folder from that repo is copied. The rest (the Claude Code plugin, Cursor rule, scripts and repo docs) is not.

`tool/sync_skills.sh` downloads a release, checks its checksum and writes the folder. It can wait for the assets, because agent-skills uploads them a little after publishing the release. CI runs it with `--check` to make sure the copied files still match that release. A Claude routine listens for agent-skills releases, runs the script and opens a PR here with the changelog entry, so there is no sync workflow in this repo.

Closes https://github.com/powersync-ja/powersync.dart/issues/472

AI use: Created and reviewed with Claude Code.


packages/powersync/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.4.1 (unreleased)
```
simolus3: This is arguably new functionality, so it should be `2.5.0`

tool/sync_skills.sh
```diff
@@ -0,0 +1,150 @@
...
+# packages/powersync/skills/, where `dart run skills@ get` finds it for users of
+# this package.
+#
+# Usage:
+#   tool/sync_skills.sh                # sync from the latest agent-skills release
+#   tool/sync_skills.sh v1.4.0         # sync from a specific release tag
+#   tool/sync_skills.sh --wait v1.4.0  # same, but wait up to 10 minutes for the
+#                                      # release assets, which agent-skills uploads
+#                                      # a little after publishing the release
+#   tool/sync_skills.sh --check        # re-sync the recorded release into a temp
```
simolus3: We don't need this script to run for minutes, we can probably just have it fail immediately (to let us retry manually). Let's remove the `--wait` option.

simolus3 (commented):
Should we skip adding files that aren't needed for Dart (like skills for other SDKs)?

PR #477, opened 2026-09-18
Move root dependencies to dev_dependencies and remove unused ones

spydon:
The root workspace package is `publish_to: none` and only used for tooling (`melos`, `lints`, and `path`/`yaml` in `tool/`), so its dependencies are now `dev_dependencies`.

Also removes dependencies that are never imported:

- `test` in the root `pubspec.yaml`
- `fake_async` in `powersync`
- `test` in `powersync_attachments_helper` (no `test/` directory)
- `flutter_test` in `powersync_devtools_extension` (no `test/` directory)

No resolved versions change in `pubspec.lock`, only the dependency kinds.

Verified locally:

- `melos run prepare`
- `dart format` and `dart analyze --fatal-infos --fatal-warnings`
- `dart test -p chrome,vm` in `packages/powersync` (all pass except the Chrome encryption smoke test, which needs `sqlite3mc.wasm` that is only built in CI)
- Encryption tests with sqlite3mc and sqlcipher
- DevTools extension build and `melos publish --dry-run`


simolus3 (approved):
Thanks Lukas!

